PR #12, opened 2025-05-28
Overriding config

### LINKED ISSUE #5: Add overriding config parameters at launch time
### Feature request

Currently, when one needs to tweak some hyper-parameters in bundled or individual model / training / data configs (e.g. `configs/bert_relpe/bert_large_rope_swiglu_seq128.json` or `configs/bert_relpe/data_seq256_bf16.json`) or DeepSpeed configs (e.g. `configs/bert_relpe/deepspeed_config_seq128_bf16.json`), they need to be manually edited in-place or after being copied. It becomes problematic when there's a pre-planned sequence of experiments to be run in succession with variation of one or several hyper-parameters. The only way to do it now is to produce several copies of the config and alter them which is non-scalable, error-prone and inconvenient. 

It would be desirable to override parameters via CLI during the launch of experiments, for example:

```
configs/lra/ds_train_dense_attn_pathfinder32.sh --model_config.embedding_dropout=0.05 
configs/lra/ds_train_dense_attn_pathfinder32.sh --model_config.local_attention=true deepspeed_config.fp16.enabled=False
configs/lra/ds_train_dense_attn_pathfinder32.sh --data.test.inputs="input/valid.src" --data.test.labels="label/test.label"
configs/lra/ds_train_dense_attn_pathfinder32.sh --model_config.embedding_dropout=0.05 training.lr_schedule="cosine" \ 
  training.lr_scheduler_params.degree=1

```
When implementing this feature, please make sure that changed parameters are properly logged in ML tracking systems (ClearML, Wandb).

I'm looking favorably to integrating OmegaConf or Hydra config system into the framework, but it's perfectly fine and, perhaps, preferable to use simpler solutions if possible. 

Examples of how other research labs manage configs in ML experiments to draw inspiration from:

https://github.com/fla-org (e.g "flame" repo)

https://github.com/state-spaces/ (e.g. S4 repo)

https://github.com/HazyResearch (e.g., "Based" repo)

### Required steps:

1) Research the potential ways to refactor or augment the framework and come up with one or few suggestions
2) Explain and discuss them, refine if necessary, get an approval for moving further with chosen solution.
3) Implement and test it

mirrtl:
Support for config overrides during script execution has been added. Overrides are specified in the format **cf.key=value** or **ds.key=value** after the script name.
For example:
**ds_train_dense_attn_pathfinder32.sh --override cf.model_config.num_attention_heads=1 cf.model_config.local_attention=false ds.train_batch_size=32 cf.data.validation.inputs="input/train.src".**

The implementation is based on the **--override** flag and further processing in train_arguments.py, which includes the functions **parse_override_string()**, **apply_overrides()**, **set_nested_checked()**, **parse_value()**, and **save_configs()**.

These functions provide safe parsing of keys in the format **cf.key=value** or **ds.key=value**, key validation, logging of all changes, and saving the updated configurations.

### DIFF
--- a/configs/lra/ds_train_dense_attn_pathfinder32.sh
+++ b/configs/lra/ds_train_dense_attn_pathfinder32.sh
@@ -28,7 +28,11 @@ if [ "${1-}" = "--resume" ]; then
     LOAD_EPOCH=""
   fi
 
-  [ $# -ge 1 ] && [ "$1" != "--override" ] || { echo "Usage: $0 --resume [EPOCH|last] JOB_NAME"; exit 1; }
+  [ $# -ge 1 ] && [ "$1" != "--override" ] || {
+    echo "Usage: $0 --resume [EPOCH|last] JOB_NAME"
+    echo "   or: $0 --resume [EPOCH|last] JOB_NAME --override cf.key=value ds.key=value ..."
+    exit 1
+  }
   SUBDIR=$1; shift
 
   if [ "${1-}" = "--override" ]; then
--- a/configs/lra/ds_train_dense_attn_pathfinder32.sh
+++ b/configs/lra/ds_train_dense_attn_pathfinder32.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+params=""
+if [ $# -ne 0 ]; then
+    params="$*"
+fi
+
 base_dir=`pwd`
 
 SEED=${SEED:-100}
@@ -70,4 +75,5 @@ NCCL_TREE_THRESHOLD=0 deepspeed --include localhost:"$NODE" --master_port "$MAST
 --load_training_checkpoint $CHECKPOINT_BASE_PATH \
 --load_checkpoint_id $CHECKPOINT_EPOCH_NAME \
 --project_name "lra-pathfinder-32" \
+--override $params \
 &> ${JOB_NAME}.log
--- a/configs/lra/ds_train_dense_attn_pathfinder32.sh
+++ b/configs/lra/ds_train_dense_attn_pathfinder32.sh
@@ -28,7 +28,7 @@ if [ "${1-}" = "--resume" ]; then
     LOAD_EPOCH=""
   fi
 
-  [ $# -ge 1 ] || { echo "Usage: $0 --resume [EPOCH|last] JOB_NAME"; exit 1; }
+  [ $# -ge 1 ] && [ "$1" != "--override" ] || { echo "Usage: $0 --resume [EPOCH|last] JOB_NAME"; exit 1; }
   SUBDIR=$1; shift
 
   if [ "${1-}" = "--override" ]; then
--- a/configs/lra/ds_train_dense_attn_pathfinder32.sh
+++ b/configs/lra/ds_train_dense_attn_pathfinder32.sh
@@ -1,10 +1,5 @@
 #!/bin/bash
 
-params=""
-if [ $# -ne 0 ]; then
-    params="$*"
-fi
-
 base_dir=`pwd`
 SEED=${SEED:-100}
 NODE=${NODE:-0}
@@ -20,6 +15,7 @@ BASE_JOB_NAME="lra_pathfinder_32"
 : "${BASE_DATA_DIR:=${base_dir}/data}"
 CHECKPOINT_BASE_PATH=""
 CHECKPOINT_EPOCH_NAME=""
+OVERRIDE_ARGS=()
 
 if [ "${1-}" = "--resume" ]; then
   shift
@@ -33,7 +29,11 @@ if [ "${1-}" = "--resume" ]; then
   fi
 
   [ $# -ge 1 ] || { echo "Usage: $0 --resume [EPOCH|last] JOB_NAME"; exit 1; }
-  SUBDIR=$1
+  SUBDIR=$1; shift
+
+  if [ "${1-}" = "--override" ]; then
+    OVERRIDE_ARGS=( "${@:2}" )
+  fi
 
   CHECKPOINT_BASE_PATH="${OUTPUT_DIR}/saved_models/${SUBDIR}"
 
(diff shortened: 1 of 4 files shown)


train_arguments.py
```diff
@@ -1,6 +1,20 @@
 import sys
 import argparse
+import json
+import logging
 
+logging.basicConfig(
```
andrewargatkiny, line 6: Use logger from `utils/logger.py` instead as it correctly handles multi-node training.

train_arguments.py
```diff
@@ -364,6 +378,93 @@ def get_argument_parser():
         default="nccl",
         help='Backend for distributed training.'
     )
-
+    parser.add_argument(
+        '--override',
+        type=str,
+        default=None,
+        nargs='+',
```
andrewargatkiny, line 385: Did you check if it works if there are no arguments to override? I think it might fail, either here or further in the code.

train_arguments.py
```diff
@@ -364,6 +378,93 @@ def get_argument_parser():
...
+        help='Override model and deepspeed configs parameters'
```
andrewargatkiny, line 386: Please add examples in the help message.

train_arguments.py
```diff
@@ -364,6 +378,93 @@ def get_argument_parser():
...
+    key_path = override[first_dot + 1:end_eq]
+    value = override[end_eq + 1:]
+    
+    logger.info(
+        "Parsed override: cfg_name=%s, key_path=%s, value=%s",
+        config_paths[cfg_name], key_path, value
+    )
+    return (cfg_name, key_path, value)
+
+def set_nested_checked(config_dict: dict, key_path: str, value):
```
andrewargatkiny, line 410: If you take some code from another repo or your code is heavily inspired by it, please reference it in a comment.

I don't know if it's true for any of the code in this PR, so it's just a reminder.

train_arguments.py
```diff
@@ -364,6 +378,93 @@ def get_argument_parser():
...
+def apply_overrides(configs: dict[str, dict], overrides: list[tuple[str, str, str]]):
+    """Apply a list of parsed overrides to their respective configuration dictionaries.
+    """
+    for cfg_name, key_path, raw_value in overrides:
+        if cfg_name not in configs:
+            raise KeyError(f"Use --override cf. or --override ds.")
+        value = parse_value(raw_value)
+        set_nested_checked(configs[cfg_name], key_path, value)
+
+def save_configs(configs: dict[str, dict], paths: dict[str, str]):
```
andrewargatkiny, line 445: The configs should not be overwritten on disk. The code state in the repo should not be altered programmatically.

train_arguments.py
```diff
@@ -364,6 +378,93 @@ def get_argument_parser():
...
+        target_dict = target_dict[key]
+
+    final_key = keys[-1]
+    if final_key not in target_dict:
+        raise KeyError(f"The final key '{final_key}' does not exist in '{key_path}'")
+
+    target_dict[final_key] = value
+    logger.info("Set %s = %r", key_path, value)
+    
+def parse_value(val: str):
```
andrewargatkiny, line 428: Please explain why do you need to decode the value as a json? Is it for inferring the type of a value?

andrewargatkiny (changes requested):

configs/lra/ds_train_dense_attn_pathfinder32.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+params=""
+if [ $# -ne 0 ]; then
+    params="$*"
```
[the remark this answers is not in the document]
mirrtl: fixed, currently --resume and --override collaboration is supported.

train_arguments.py
mirrtl: Changed to use print instead.

train_arguments.py
mirrtl: Fixed, currently using nargs="*" with default=[].

train_arguments.py
mirrtl: This implementation is inspired by an idea from another repository, but the implementation, approach, and code are different.

train_arguments.py
mirrtl: This is used for correctly inserting values into nested dicts.

train_arguments.py
```diff
@@ -379,6 +381,90 @@ def get_argument_parser():
...
+    config_paths = {
+        "cf": args.config_file,
+        "ds": args.deepspeed_config
+    }
+    configs = {}
+    
+    for name, path in config_paths.items():
+        with open(path, 'r', encoding='utf-8') as f:
+            configs[name] = json.load(f)
+            print("Loaded config '{name}' from {path}")
```
andrewargatkiny, line 461: This flow (invocation of function `override_configs`) is conditional but you instantiate configs here. Specifically, `deepspeed_config` is instantiated *only * in this function as you commented it out in the other function, and `config` is instantiated in two places which could add confusion and cause errors in future code refactors.

Additionally, code outside of this function which succeeds its invocation, can potentially overwrite back the changes produced by it (see lines 589-503). This is unintended behavior as I envisioned the override params from CLI to act as a final ruling.

Therefore, I recommend that you let the configs be instantiated in the code as they originally were, then wait for all changes to the configs made in the code to happen, and then invoke this function. It should just change the configs in place, not loading them from drive for the second time.

train_arguments.py
```diff
@@ -379,6 +381,90 @@ def get_argument_parser():
...
+        help='Override parameters for config files. Accepts values in the form '
+             '"cf.key=val" or "ds.key=val" to override nested values in the model (cf) '
+             'or deepspeed (ds) configs. For nested keys, use dot notation: '
+             'key = key1.key2.key3.'
+        
+    )
 
     return parser
+
+def parse_override_string(config_paths: dict, override: str):
```
andrewargatkiny, line 398: Argument `config_paths` is redundant in this function except for a log message which would be more informative showing the real cfg_name, so I suggest to remove it.

train_arguments.py
andrewargatkiny: This way, you don't need `config_paths` variable at all.

andrewargatkiny (changes requested):

andrewargatkiny (changes requested):

andrewargatkiny (approved):

Issue #6, opened 2025-05-06
Create adapter for HuggingFace models


andrewargatkiny:
## Feature request

There should be an option to use any architecture from HuggingFace (e.g. https://huggingface.co/docs/transformers/main/en/model_doc/nystromformer#transformers.NystromformerModel) or HF-compatible model (e. g. https://github.com/fla-org/flash-linear-attention/blob/main/fla/models/gla/modeling_gla.py#L167) for training and evaluation within the existing framework scope.

Tentative guidelines on how to accomplish it:

- Create a mechanics which receives a config (Python dictionary) and, based on `model_name` (or it can be named `model_arch`) property in the dictionary, instantiates relevant model config class, populating it with other values from the dictionary. Then it initiates the relevant model class with the config.
- For all HF-compatible architectures, you need the basic "trunk" of the model, without any language modeling or other heads. Create an adapter class which is able to wrap the model and expose the same `forward ` interface as in https://github.com/andrewargatkiny/dense-attention/blob/518f74a6bd34a3eb120feb8b21de109a79c383f7/src/other_models/modeling.py#L889 or https://github.com/andrewargatkiny/dense-attention/blob/518f74a6bd34a3eb120feb8b21de109a79c383f7/src/modeling.py#L404, converting inputs and outputs when necessary. 
- Adapt or copy verbatim other classes for language modeling, classification, regression, etc.
- Create and register [`Task`](https://github.com/andrewargatkiny/dense-attention/blob/518f74a6bd34a3eb120feb8b21de109a79c383f7/utils/tasks.py) classes for HF-compatible models
- Write tests and launch them to ensure everything works

> WARNING: DANet, my implementation of vanilla and linear Transformers don't support auto-regressive generation with KV cache yet, so it's not required to integrate generation interfaces for HF-compatible models as of this moment.

Please feel free to discuss and consult any details in this thread before actually implementing this feature.

PR #14, opened 2025-06-12
Add support for HuggingFace models

herbitovich:
Adapter classes for HuggingFace models. Implements #6


utils/tasks.py
```diff
@@ -187,6 +188,37 @@ class GluePretrainingMLM:
...
+class HFBertPretraining:
+    dataset_type = BertPretrainingDatasetFactory
+    model_type = HFForPreTraining
+    eval_func = eval_mlm_classification_task
+
+@dataclass
+class HFGlueForRegression:
+    dataset_type = GlueBertDataset
+    model_type = HFForRegression
+    eval_func = eval_regression_task
```
andrewargatkiny, line 222: Not all tasks like BertMLM and GPT are replicated for HF models.

utils/tasks.py
```diff
@@ -229,3 +261,8 @@ def get_task(cls, name):
 TaskRegistry.register_task("glue_hf_with_acc_metrics", GlueHFWithAccMetrics)
 TaskRegistry.register_task("glue_transformer_with_all_metrics", GlueTransformerWithAllMetrics)
 TaskRegistry.register_task("glue_transformer_for_regression", GlueTransformerForRegression)
+TaskRegistry.register_task("hf_sequence_classification", HFSequenceClassification)
+TaskRegistry.register_task("hf_text_classification", HFTextClassification)
+TaskRegistry.register_task("hf_texts_matching", HFTextsMatching)
+TaskRegistry.register_task("hf_bert_pretraining", HFBertPretraining)
+TaskRegistry.register_task("hf_glue_for_regression", HFGlueForRegression)
```
andrewargatkiny, line 268: Not all tasks are registered.

src/model_config.py
```diff
@@ -201,6 +204,18 @@ def __init__(self,
             self.hybrid = hybrid
             self.transformer_heads = transformer_heads
             self.attn_proj_biases = attn_proj_biases
+            self.hf_config = None
+            if hf_model_name_or_path:
+                self.hf_config = AutoConfig.from_pretrained(hf_model_name_or_path)
+                self.hidden_size = self.hf_config.hidden_size
```
andrewargatkiny, line 210: hf_config values should be populated from the user-supplied config, not the other way around. It should also always be checked if the hf_config has the same parameters as in user-supplied config, and throw exceptions otherwise.

andrewargatkiny (changes requested):

src/model_config.py
```diff
@@ -51,7 +52,9 @@ def __init__(self,
                  local_scheme="l_sl_g",
```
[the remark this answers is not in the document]
andrewargatkiny: Its initialization function should have only `hf_model_name_or_path` and `**kwargs` arguments, and then each of `**kwargs` should replace the corresponding value in `hf_config`.

herbitovich:
@andrewargatkiny in configs/lra_mlm, task scripts and configs (e.g. ds_train_dense_attn_mlm_cls_listops.sh & dense_attn_listops.json) are outdated: lr_schedule and lr_offset are passed as train arguments, but in the current version of the framework they are supposed to be placed in config["training"]. also, there are issues with path variables and use of the deprecated --use-nvidia-dataset, --use-local-attention args in the scripts

deepspeed_train.py
```diff
@@ -313,7 +315,7 @@ def train(args,
 def update_weights_scalers(model, num_layers):
     """Update weights scalers of DenseAttention Model"""
     for i in range(num_layers):
-        ffn = model.bert.encoder.layer[i].ffn
+        ffn = attrgetter(model.path_to_bert)(model).encoder.layer[i].ffn
```
andrewargatkiny: We need as much as possible abstracted path – i.e. in this case it should use not just model.PATH_TO_BERT but model.PATH_TO_LAYERS and FFN_NAME where PATH_TO_LAYERS="bert.encoder.layer" and FFN_NAME="ffn".

Reason: in different models, their parameter names can differ.

deepspeed_train.py
```diff
@@ -648,30 +657,30 @@ def prepare_optimizer_parameters(args, model):
         weight_decay = 0.01
 
 
-    groups = [{'params': list(model.bert.embeddings.parameters()),
-               'lr': 0.0,
-               'weight_decay': weight_decay,
-               'name': 'embeddings'}]
-    for i in range(len(model.bert.encoder.layer)):
+    groups = [{'params': list(attrgetter(model.path_to_bert)(model).embeddings.parameters()),
```
andrewargatkiny: Please either 1) revert lines 660-687 to the original state and comment them out using triple quotes (no need to delete as they might be useful in the future), or  2) make proper abstract paths as described in L318.

deepspeed_train.py
```diff
@@ -704,7 +713,7 @@ def prepare_model_optimizer(args):
     deepspeed.init_distributed(dist_backend=args.dict_backend)
     args.local_rank = int(os.environ['LOCAL_RANK'])
     model_class = args.task.model_type
-    config_class = ModelConfig
+    config_class = HFConfig if 'hf_model_name_or_path' in args.config["model_config"] else ModelConfig
```
andrewargatkiny: Please revert this line to previous state. Instead, in all types of tasks for HF models, assign config_type = HFConfig. Then it will be automatically picked up in lines 717-718

utils/tasks.py
```diff
@@ -187,7 +188,82 @@ class GluePretrainingMLM:
...
+class HFTextsMatching:
+    dataset_type = AANDataset
+    model_type = HFForAANMatching
+    eval_func = eval_classification_task
+
+@dataclass
+class HFBertPretraining:
+    dataset_type = BertPretrainingDatasetFactory
+    model_type = HFForPreTraining
+    eval_func = eval_mlm_classification_task
```
andrewargatkiny, line 218: See comment for line 716.

src/modeling.py
```diff
@@ -41,6 +41,19 @@
 
 logger = logging.getLogger(__name__)
 
+def get_num_params(model, non_embedding=True):
```
andrewargatkiny: As all the architectures use this function, it's better to move it to some "neutral territory" – `train_utils.py` would be a good choice.

src/modeling.py
```diff
@@ -586,6 +585,7 @@ class DANetForPreTraining(DANetPreTrainedModel):
 
     def __init__(self, config: ModelConfig, args):
         super(DANetForPreTraining, self).__init__(config)
+        self.path_to_bert = "bert"
```
andrewargatkiny: Make the variable name all uppercase to indicate that it's a constant string.
Also change it to "PATH_TO_BACKBONE".

andrewargatkiny: Pro tip: if you need to store the paths to some model parts which are inside the backbone, it's better to store them *inside* the backbone class – i.e. `DANetModel` in this case.

deepspeed_train.py
andrewargatkiny: Specifically, you need to define `PATH_TO_LAYERS` and `PATH_TO_EMBEDDINGS` for all existing model types in the framework (DANet, Transformer, HFmodel) and use them to refactor the code in  `report_model_weights` function.

src/other_models/hf_modeling.py
```diff
@@ -0,0 +1,182 @@
+# src/hf_modeling.py
+from transformers import AutoModel, AutoConfig
+import torch
+import torch.nn as nn
+from ..model_config import ModelConfig
+from src.modeling import *
+import inspect
+
+class HFConfig(ModelConfig):
```
andrewargatkiny: It seems that there's no need for `ModelConfig` at all. You can delete all the code related to it. 

If you found any specific reasons why the config for DANet should be instantiated here, please provide them.

andrewargatkiny: If there are some attributes of the config being accessed by outside code, you can   define `__getattr__` or `__getattribute__` methods in this class so it redirects all  call to inner `hf_config`'s attributes.

src/other_models/hf_modeling.py
```diff
@@ -0,0 +1,182 @@
...
+        pooled_output = sequence_output[:, 0]
+        
+        return sequence_output, pooled_output
+
+# Task-Specific Heads
+class HFForPreTraining(DANetForPreTraining):
+    def __init__(self, config, args=None):
+        super().__init__(config, args)
+        self.path_to_bert = "bert.model"
+        self.bert = HFBaseModel(config)
```
andrewargatkiny: here and in the other places in this file, please change `bert` to `backbone`.

utils/tasks.py
andrewargatkiny: If you resolve a previous comment, please always write "fixed" or "done".

utils/tasks.py
andrewargatkiny: Likewise, have you taken any action about this feedback?

src/other_models/hf_modeling.py
```diff
@@ -0,0 +1,182 @@
...
+
+        super().__init__(**config_kwargs)
+        self.hf_config = AutoConfig.from_pretrained(hf_model_name_or_path)
+        for param, value in kwargs.items():
+            if param == "vocab_size_or_config_json_file" and "vocab_size" in self.hf_config:
+                setattr(self.hf_config, "vocab_size", value) 
+                continue
+            setattr(self.hf_config, param, value) 
+
+class HFBaseModel(DANetPreTrainedModel):
```
andrewargatkiny: In all model class definitions in this file you inherent from base classes which are not imported.
However, the HF classes should not inherent from existing DANet or Transformer classes at all. Please see my previous comments (outdated, so look in "conversation" section) about a proper base class which should hold relevant init function for any architecture.

andrewargatkiny: I also don't see any call for weight initialization in any of the new model classes.

src/other_models/hf_modeling.py
```diff
@@ -0,0 +1,182 @@
...
+    """Core adapter that handles input/output conversion"""
+    def __init__(self, config, args=None):
+        super().__init__(config)
+        self.hf_config = config.hf_config
+        self.model = AutoModel.from_config(self.hf_config)
```
andrewargatkiny: You can investigate if this `AutoModel` already inherits the base class with weight init logic in case of any HF architecture.

configs/bert/ds_train_hf_bert.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+base_dir=`pwd`
+OUTPUT_DIR=${base_dir}/bert_model_dense_attn_adam_outputs
+BASE_JOB_NAME="hf_bert_pretraining"
```
andrewargatkiny, line 8: There's no need for glue and lra_mlm configs for hf models, at least for now. Move all lra hf configs to dedicated sub-dir within `configs/lra` dir.

andrewargatkiny (changes requested):

configs/bert/ds_train_hf_bert.sh
herbitovich: done

utils/tasks.py
herbitovich: done

utils/tasks.py
herbitovich: done

deepspeed_train.py
herbitovich: fixed by adding the `PATH_TO_LAYERS` and `PATH_TO_EMBEDDINGS` properties to base classes (HFAdapter, DANetModel, BertModel)

deepspeed_train.py
herbitovich: done (made paths abstract)

deepspeed_train.py
herbitovich: fixed

utils/tasks.py
herbitovich: fixed, added config_type for every hf task.

src/modeling.py
herbitovich: done

src/modeling.py
herbitovich: fixed. path_to_backbone is now deprecated as all models refer to self.bert as their backbone now. groups of parameters (embeddings, layers) are indexed through corresponding paths in `PATH_TO_EMBEDDING` and `PATH_TO_LAYERS`

src/other_models/hf_modeling.py
herbitovich: fixed the issue by making HFConfig an independent class. The __getattr__ method is not necessary because all passed kwargs are assigned to the class object itself, rather than only to self.hf_config.

src/other_models/hf_modeling.py
herbitovich: fixed, removed it altogether

src/other_models/hf_modeling.py
herbitovich: fixed. base class & weight initialization for each model has been added

src/other_models/hf_modeling.py
herbitovich: yes, each model class has an _init_weights method that applies specific initialization rules

deepspeed_train.py
```diff
@@ -441,8 +443,16 @@ def report_model_activations(args, model, data, step, bins=20, **kwargs):
...
             def hook(model, input, output):
-                if not isinstance(output, (list, tuple)):
-                    activations[name] = output.cpu().float().numpy()
+                if isinstance(output, torch.Tensor):
+                    tensor_output = output
+                elif hasattr(output, 'last_hidden_state'):
+                    tensor_output = output.last_hidden_state
+                elif isinstance(output, (tuple, list)):
+                    for item in output:
+                        if isinstance(item, torch.Tensor):
```
andrewargatkiny, line 452: Please give an example of what could be such a tensor?

deepspeed_train.py
```diff
@@ -504,28 +513,31 @@ def report_model_weights(args, model, step, bins=20):
...
-                layer = layer.group(0)
-                group_name = name.replace(layer, '.', 1)
-                layer_number = re.search(r'\d+', layer).group(0)
-                return group_name, f"Layer {layer_number}"
-
-            return ("Embeddings parameters", name) if re.search(r'embed', name) else ("Other parameters", name)
+        backbone = attrgetter(model.PATH_TO_BACKBONE)(model)
+        layers = attrgetter(backbone.PATH_TO_LAYERS)(backbone)
+        embeddings = attrgetter(backbone.PATH_TO_EMBEDDINGS)(backbone)
+        full_path = lambda name, is_layer=True: \
```
andrewargatkiny: Please add the comments detailing the purpose and the logic of the code (what it accomplishes and how it works).

andrewargatkiny: Regarding L519, it seems it can be merged with the next line without exceeding the characters limit.

deepspeed_train.py
```diff
@@ -504,28 +513,31 @@ def report_model_weights(args, model, step, bins=20):
...
+            model.PATH_TO_BACKBONE, 
+            (backbone.PATH_TO_LAYERS if is_layer else backbone.PATH_TO_EMBEDDINGS),
+            name])
+        embeddings_params = {
+            full_path(name, is_layer=False): ('Embedding parameters', name) 
+            for name, param in 
+            embeddings.named_parameters()
+        }
+        layers_params = {
+            full_path(name): ('.'.join(name.split(".")[1:]),
```
andrewargatkiny, line 535: It would be nice to write a comment with a possible example of the `layers_params` dictionary entry.

deepspeed_train.py
```diff
@@ -647,31 +659,33 @@ def prepare_optimizer_parameters(args, model):
...
             # params, they wouldn't be updated.
             groups.append({
-                'params': list(model.bert.encoder.layer[i].attention.parameters()),
+                'params': list(layers[i].attention.parameters()),
                 'lr': 0.0,
                 'weight_decay': weight_decay,
                 'name': f'layer_{i}_attention'
             })
-            if hasattr(model.bert.encoder.layer[i], 'ffn'):
+            if hasattr(layers, 'ffn'):
```
andrewargatkiny: layers[i]

deepspeed_train.py
```diff
@@ -647,31 +659,33 @@ def prepare_optimizer_parameters(args, model):
...
                 groups.append({
-                    'params': list(model.bert.encoder.layer[i].ffn.parameters()),
+                    'params': list(layers.ffn.parameters()),
```
andrewargatkiny: layers[i]

deepspeed_train.py
```diff
@@ -796,15 +810,17 @@ def run(args, model, optimizer, start_epoch):
     config = args.config
     logger = args.logger
     task = args.task
+    backbone = attrgetter(model.PATH_TO_BACKBONE)(model)
```
andrewargatkiny: Please move L813-814 inside `if` block.

src/other_models/hf_modeling.py
```diff
@@ -0,0 +1,202 @@
...
+    def __init__(self, hf_model_name_or_path, **kwargs):
+        self.hf_config, unused_kwargs = AutoConfig.from_pretrained(hf_model_name_or_path, **kwargs, return_unused_kwargs=True)
+        for param, value in unused_kwargs.items():
+            setattr(self, param, value)
+    def __getattribute__(self, name):
+        try:
+            return object.__getattribute__(self, name)
+        except AttributeError:
+            pass
+        return getattr(self.hf_config, name)
```
andrewargatkiny, line 17: Why did you decide to put this line here, not in the `except` block?

src/other_models/hf_modeling.py
```diff
@@ -0,0 +1,202 @@
...
+            if module.bias is not None:
+                module.bias.data.zero_()
+
+class HFAdapter(nn.Module):
+    """Core adapter that handles input/output conversion"""
+    def __init__(self, config, args=None):
+        super().__init__()
+        self.config = config
+        self.model = AutoModel.from_config(self.config.hf_config)
+        self.PATH_TO_LAYERS = "model.encoder.layer"
```
andrewargatkiny: Please make `PATH_TO_LAYERS` and `PATH_TO_EMBEDDINGS` load from the `config` object, and place these values ("model.encoder.layer" and "model.embeddings") as the defaults for the corresponding config attributes.

src/other_models/hf_modeling.py
```diff
@@ -0,0 +1,202 @@
...
+        self.PATH_TO_EMBEDDINGS = "model.embeddings"
+
+    def forward(self, input_ids, attention_mask=None, token_type_ids=None):
+        input_ids = input_ids[:, :self.config.max_position_embeddings]
```
andrewargatkiny: If we delete this constraint on `max_position_embeddings` and supply a sequence with length exceeding it, will it break the algorithm in the inner HuggingFace model?

If not, I'd like for this constraint to be removed, because there would be many cases when we possibly need to check greater sequence lengths than those the model was trained on.

configs/bert/ds_train_hf_bert.sh
andrewargatkiny: `glue` and `lra_mlm` hf configs are still present. Please move them to dedicated directories or remove completely.

andrewargatkiny (dismissed):

src/other_models/hf_modeling.py
herbitovich: This way, if a non-existent attribute name is passed as an argument, the error message doesn't become nested. Take a look at the example below:
```python
def foo(x):
    try:
        return 5/x
    except ZeroDivisionError:
        return 4/x
```    

```
>>> foo(0)
Traceback (most recent call last):
  File "<python-input-0>", line 3, in foo
    return 5/x
           ~^~
ZeroDivisionError: division by zero

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    foo(0)
    ~~~^^^
  File "<python-input-0>", line 5, in foo
    return 4/x
           ~^~
ZeroDivisionError: division by zero
```

**But, if we pass the first exception catch:**
```python
def foo(x):
    try:
        return 5/x
    except ZeroDivisionError:
        pass
    return 4/x
```
```
>>> foo(0)
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    foo(0)
    ~~~^^^
  File "<python-input-2>", line 6, in foo
    return 4/x
           ~^~
ZeroDivisionError: division by zero
```

src/other_models/hf_modeling.py
herbitovich: It will, unfortunately.
```
[rank0]: RuntimeError: The size of tensor a (1024) must match the size of tensor b (24) at non-singleton dimension 1
```

deepspeed_train.py
herbitovich: Preliminary outputs of encoder.layers. The following debug line:
```python
elif isinstance(output, (tuple, list)):
    print(f"DEBUG_TUPLE_LIST_HOOK: Module '{name}' output is {type(output)} with {len(output)} elements")
    for item in output:
        ...
```
Resulted in this:
```
DEBUG_TUPLE_LIST_HOOK: Module 'module.backbone.model.encoder.layer.0.attention.self' output is <class 'tuple'> with 1 elements
DEBUG_TUPLE_LIST_HOOK: Module 'module.backbone.model.encoder.layer.0.attention' output is <class 'tuple'> with 1 elements
DEBUG_TUPLE_LIST_HOOK: Module 'module.backbone.model.encoder.layer.0' output is <class 'tuple'> with 1 elements
DEBUG_TUPLE_LIST_HOOK: Module 'module.backbone.model.encoder.layer.1.attention.self' output is <class 'tuple'> with 1 elements
...
```
The tuple just wraps a tensor:
```
(tensor([[[-0.0370,  0.0774, -0.0013,  ..., -0.0945,  0.0536,  0.0308],
         [-0.0383,  0.0617, -0.0031,  ..., -0.0856,  0.0529,  0.0367],
         [-0.0327,  0.0746, -0.0031,  ..., -0.0850,  0.0667,  0.0356],
         ...,
         [-0.0400,  0.0703, -0.0023,  ..., -0.0945,  0.0637,  0.0353],
         [-0.0410,  0.0776, -0.0027,  ..., -0.0889,  0.0656,  0.0340],
         [-0.0410,  0.0688, -0.0046,  ..., -0.0833,  0.0629,  0.0339]],

        [[-0.0348,  0.0632,  0.0013,  ..., -0.0839,  0.0623,  0.0317],
         [-0.0323,  0.0632,  0.0024,  ..., -0.0945,  0.0596,  0.0403],
         [-0.0344,  0.0684,  0.0024,  ..., -0.0931,  0.0554,  0.0385],
         ...,
         [-0.0212,  0.0602,  0.0026,  ..., -0.0928,  0.0520,  0.0446],
         [-0.0327,  0.0729,  0.0122,  ..., -0.0884,  0.0667,  0.0341],
         [-0.0188,  0.0728,  0.0041,  ..., -0.0870,  0.0568,  0.0396]],

        [[-0.0416,  0.0707, -0.0030,  ..., -0.0782,  0.0698,  0.0281],
         [-0.0416,  0.0674, -0.0094,  ..., -0.0880,  0.0635,  0.0288],
         [-0.0387,  0.0598, -0.0056,  ..., -0.0865,  0.0682,  0.0312],
         ...,
         [-0.0297,  0.0652, -0.0042,  ..., -0.0887,  0.0730,  0.0313],
         [-0.0412,  0.0410, -0.0053,  ..., -0.0826,  0.0575,  0.0313],
         [-0.0401,  0.0706, -0.0057,  ..., -0.0855,  0.0698,  0.0339]],

        ...,

        [[-0.0383,  0.0703, -0.0038,  ..., -0.0800,  0.0525,  0.0270],
         [-0.0305,  0.0740, -0.0080,  ..., -0.0872,  0.0526,  0.0341],
         [-0.0453,  0.0734, -0.0008,  ..., -0.0967,  0.0663,  0.0341],
         ...,
         [-0.0417,  0.0783, -0.0047,  ..., -0.0842,  0.0612,  0.0264],
         [-0.0363,  0.0706, -0.0049,  ..., -0.0994,  0.0623,  0.0306],
         [-0.0421,  0.0751, -0.0044,  ..., -0.0955,  0.0557,  0.0291]],

        [[-0.0385,  0.0707, -0.0024,  ..., -0.0919,  0.0666,  0.0367],
         [-0.0370,  0.0641, -0.0022,  ..., -0.0822,  0.0754,  0.0385],
         [-0.0368,  0.0696, -0.0056,  ..., -0.0834,  0.0707,  0.0369],
         ...,
         [-0.0385,  0.0662,  0.0013,  ..., -0.0758,  0.0695,  0.0361],
         [-0.0309,  0.0639, -0.0082,  ..., -0.0783,  0.0712,  0.0333],
         [-0.0347,  0.0607, -0.0051,  ..., -0.0787,  0.0615,  0.0406]],

        [[-0.0407,  0.0626, -0.0063,  ..., -0.0837,  0.0605,  0.0394],
         [-0.0306,  0.0760, -0.0080,  ..., -0.0730,  0.0619,  0.0366],
         [-0.0439,  0.0782, -0.0098,  ..., -0.0887,  0.0612,  0.0373],
         ...,
         [-0.0439,  0.0773, -0.0097,  ..., -0.0894,  0.0657,  0.0305],
         [-0.0370,  0.0585, -0.0091,  ..., -0.0755,  0.0467,  0.0416],
         [-0.0411,  0.0547, -0.0130,  ..., -0.0947,  0.0550,  0.0395]]],
       device='cuda:0', dtype=torch.float16),)
```

src/other_models/hf_modeling.py
herbitovich: done

deepspeed_train.py
herbitovich: done

deepspeed_train.py
herbitovich: fixed

deepspeed_train.py
herbitovich: fixed

deepspeed_train.py
herbitovich: layers_params is analogical to embeddings_params, added a comment with an example content of the `params` dictionary, which contains both.

deepspeed_train.py
herbitovich: done

herbitovich: done

src/other_models/hf_modeling.py
andrewargatkiny: I assume that you mean non-existent among  both outer attributes and inner `hf_config` ones. With this in mind, it’s a great explanation!

src/other_models/hf_modeling.py
andrewargatkiny: Got it. Nevertheless, I’d like for this exception to be explicit rather than silently truncating long sequences, since it may lead to wrong analysis and conclusions. So, please delete the truncation here and below.

deepspeed_train.py
```diff
@@ -504,28 +513,43 @@ def report_model_weights(args, model, step, bins=20):
...
+            full_path(name, is_layer=False): ('Embedding parameters', name) 
+            for name, param in 
+            embeddings.named_parameters()
+        }
+        layers_params = {
+            full_path(name): ('.'.join(name.split(".")[1:]),
+                              f'Layer {name.split(".")[0]}')
+            for name, param in layers.named_parameters()
+        }
+        """
```
andrewargatkiny, line 539: Please add a comment here “Example dictionary entry”.

herbitovich: done

andrewargatkiny (approved):

PR #16, opened 2025-07-01
feat: --reuse_job_name flag

glaucomaa:
by default, it writes to a new folder, but if the --reuse_job_name flag is passed, it will write to the same folder

### DIFF
--- a/configs/lra/ds_train_dense_attn_pathfinder32.sh
+++ b/configs/lra/ds_train_dense_attn_pathfinder32.sh
@@ -11,6 +11,15 @@ TRACKING_SYSTEM=${TRACKING_SYSTEM:-clearml}
 OUTPUT_DIR=${base_dir}/bert_model_dense_attn_adam_outputs
 BASE_JOB_NAME="lra_pathfinder_32"
 
+REUSE_JOB_NAME=false
+for arg in "$@"; do
+  if [ "$arg" = "--reuse_job_name" ]; then
+    REUSE_JOB_NAME=true
+    set -- "${@/--reuse_job_name/}"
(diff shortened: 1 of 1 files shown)


andrewargatkiny:
@CodiumAI-Agent /improve

configs/lra/ds_train_dense_attn_pathfinder32.sh
```diff
@@ -52,9 +72,15 @@ if [ "${1-}" = "--resume" ]; then
 
 
   JOB_NAME="${SUBDIR}_from_epoch_${LOAD_EPOCH}${JOB_NAME_SUFFIX}"
+  if $REUSE_JOB_NAME; then
+    JOB_NAME="$SUBDIR"
+  fi
 else
   # Set up for initial training
   JOB_NAME="${BASE_JOB_NAME}${JOB_NAME_SUFFIX}"
+  if $REUSE_JOB_NAME; then
```
andrewargatkiny: This would be an unintended behavior. If we don't want to have a job suffix for a fresh training run, we can just type `JOB_NAME_SUFFIX=""` before invoking the training script.

andrewargatkiny (changes requested):

configs/lra/ds_train_dense_attn_pathfinder32.sh
andrewargatkiny, line 1: Now you seem to have deleted the changes which represented the substance of this PR.

andrewargatkiny (changes requested):

andrewargatkiny (approved):

Issue #13, opened 2025-06-09
Incorporate S4 line of models


andrewargatkiny:
It's essential to incorporate a seminal line of work on State Space Models (S4, S4D and DSS – https://github.com/state-spaces/s4/tree/main/models) into this framework. Later, I'd like to also add S5 model (https://github.com/lindermanlab/S5).

They should be adapted with seamless compatibility with existent workflows for training and evaluation – e.g. using DeepSpeed, the same configs and launch scripts machinery, ClearML logging, multi-node training, etc. If the original code does not support CPU training/ inference, no need to add support for it in our framework.

Of possible relevance for this task is the repo: https://github.com/IdoAmos/not-from-scratch. However, keep in mind that PyTorch Lightning should not be used as we use DeepSpeed over plain PyTorch.

As a proof of correct work of new models, I request to also write configs for experiments with bidirectional SSMs on Pathfinder-32 and BERT pretraining, and to test that everything works. 

Don't hesitate to ask questions or consult me at any stage of the project development.

andrewargatkiny:
#35 and, to some degree, #36 are blocking the S4 integration for now, although it's possible to add support for differential learning rates and weight decays later.

PR #17, opened 2025-07-24
S4 incorporation

Rozhkin-Aleksandr:
This pull request introduces the integration of the S4, S4D, and DSS State Space Models into our framework, resolving issue #13. The implementation ensures seamless compatibility with our existing training and evaluation workflows.

**Key Changes & Features:**

*   **Core Model Implementation:**
    *   A new `s4.py` file provides a comprehensive realization of the S4 block, incorporating all necessary modifications.
    *   The `modeling_s4.py` file defines the core model classes for the sequence classification and pretraining.
*   **New Tasks & Configurations:**
    *   `tasks.py` has been updated to include new tasks for S4 model experiments in sequence classification and Masked Language Modeling (MLM) pretraining.
    *   New configurations for S4, S4D, and DSS models have been written for Pathfinder-32 and C4 pretraining (supporting bidirectional SSMs as requested).
*   **Experiments:**
    *   The newly integrated models have been thoroughly tested with the provided configurations to confirm correct functionality and adherence to framework standards.


configs/bert/deepspeed_config_seq128.json
```diff
@@ -21,7 +21,7 @@
   "wall_clock_breakdown": false,
 
   "fp16": {
-    "enabled": true,
+    "enabled": false,
```
andrewargatkiny, line 24: Please revert to the previous value.

src/other_models/s4.py
```diff
@@ -0,0 +1,1576 @@
+"""Standalone version of Structured (Sequence) State Space (S4) model."""
```
andrewargatkiny, line 1: Please reference the original code (and/ or papers) upon which your code is based, and its license, if applicable.

src/other_models/s4.py
```diff
@@ -0,0 +1,1576 @@
...
+import math
+import numpy as np
+from scipy import special as ss
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from einops import rearrange, repeat
+import opt_einsum as oe
+
+contract = oe.contract
```
andrewargatkiny, line 14: Why did you decide to use `opt_einsum` instead of `torch.einsum` like in the original code? Did you measure that it's faster?

Also, [torch.einsum](https://docs.pytorch.org/docs/stable/generated/torch.einsum.html) uses opt_einsum if its installed, so I suggest to proceed with the original code if it's not slower.

src/other_models/s4.py
```diff
@@ -0,0 +1,1576 @@
...
+    """Initializes multi-GPU-friendly python logger."""
+
+    logger = logging.getLogger(name)
+    logger.setLevel(level)
+
+    # this ensures all logging levels get marked with the rank zero decorator
+    # otherwise logs would get multiplied for each GPU process in multi-GPU setup
+ 
+    return logger
+log = get_logger(__name__)
```
andrewargatkiny, line 28: We already have a logger with correct multi-GPU handling in the project. Please consider using it instead.

src/other_models/modeling_s4.py
```diff
@@ -0,0 +1,592 @@
...
+        For non-embedding count (default), the position embeddings get subtracted.
+        The token embeddings would too, except due to the parameter sharing these
+        params are actually used as weights in the final layer, so we include them.
+        """
+        n_params = sum(p.numel() for p in self.parameters())
+        if non_embedding:
+            n_params -= self.embeddings.word_embeddings.weight.numel()
+        return n_params
+
+    def init_bert_weights(self, module):
```
andrewargatkiny, line 285: Please check that this function affects none of the parameters of the `S4 layer` module, because it has very intricate initialization having a key impact on its performance.

andrewargatkiny: Update: found at least one such case.

src/other_models/s4.py
andrewargatkiny: Update: my previous comment is not relevant since this code copies another implementation of S4, not the one I thought of.

src/other_models/s4.py
```diff
@@ -0,0 +1,1576 @@
...
+        *[ssm(measure, N, R, S // len(measures), **ssm_args) for measure in measures]
+    )
+    w = torch.cat(w, dim=0) # (S N)
+    P = torch.cat(P, dim=1) # (R S N)
+    B = torch.cat(B, dim=0) # (S N)
+    V = torch.cat(V, dim=0) # (S N N)
+    return w, P, B, V
+
+
+class OptimModule(nn.Module):
```
andrewargatkiny, line 535: This functionally wouldn't correctly work with out framework. And not only this one, but many architectures and models would benefit from granular learning rates and weight decay values per module/parameter.

It needs a thoughtful consideration on how to properly implement it. I suggest that a config for a particular layer type should contain a dict with custom lr/ wd values for some of its parameters.

Please implement this feature in a standalone PR.

src/other_models/s4.py
```diff
@@ -0,0 +1,1576 @@
...
+
+        See the class SSKernel for the kernel constructor which accepts kernel_args. Relevant options that are worth considering and tuning include "mode" + "measure", "dt_min", "dt_max", "lr"
+
+        Other options are all experimental and should not need to be configured
+        """
+
+        super().__init__()
+        self.d_model = config.d_model
+        self.d_state = config.d_state
+        self.num_hidden_layers = config.num_hidden_layers
```
andrewargatkiny, line 1388: I believe this parameter is redundant here. Are there any other unused params in this initialization?

src/other_models/s4.py
andrewargatkiny: Also cite this useful manual which details hyperparams for this implementation:
https://github.com/IdoAmos/not-from-scratch/tree/main/src/models/s4

src/other_models/modeling_s4.py
andrewargatkiny, line 1: This code is pretty much repeating `src/other_models/modeling.py`. As well as the new tasks in `utils/tasks.py`. This underlines a deeper problem: whether each new sequence-mixer/ architecture layer type should have its own outer model layers for different tasks. I'm proposing to implement a single interface which could accommodate many different sequence-mixers – see #35. 

That way, incorporating a new sequence-mixer architecture into our framework would be as simple as adding its source code file(s) and writing a few lines of harness to register the new layer type and its config type.

For now, please postpone working on this PR until #35 is finished, and then use that new feature to incorporate S4 layer into the framework.

andrewargatkiny (changes requested):

PR #18, opened 2025-07-24
Adding option of applying relpe after feature map into linear attention kernel

### LINKED ISSUE #22: Add option to apply RelPE after feature maps for Linear Transformers
Feature request: to add option to apply RelPE after feature maps for Linear Transformers. Now relative embeddings can be applied only *before* the attention kernel calculation stage, as in standard Transformers, and it would be beneficial to test it both ways.

Rozhkin-Aleksandr:
New parameter `apply_relpe_after` was added to transformer config.
Selection of the point of applying relpe was implemented.
New configs for linear attention on lra pathfinder32 were added.


configs/lra/linear_transformer/deepspeed_config_pathfinder32.json
```diff
@@ -0,0 +1,37 @@
+{
+  "train_batch_size": 100,
+  "train_micro_batch_size_per_gpu": 100,
+  "steps_per_print": 100,
+  "prescale_gradients": false,
```
andrewargatkiny: Please delete all configs

andrewargatkiny (changes requested):

src/other_models/modeling.py
```diff
@@ -330,7 +335,7 @@ def forward(self, hidden_states, attention_mask, rope_cache):
         #attention_mask = None
         context_layer = self.attention_kernel(
             query_layer, key_layer, value_layer, attn_mask=attention_mask,
-            dropout_p=self.dropout_prob, causal=self.causal
+            dropout_p=self.dropout_prob, causal=self.causal, rope_cache=rope_cache
```
andrewargatkiny, line 338: Please also consider `BertSelfLocalAttention` and `BertSelfShiftedLocalAttention` classes to also provide this feature.

src/other_models/attention_kernels.py
```diff
@@ -1,6 +1,8 @@
 import torch
 import torch.nn as nn
 from torch.nn.attention.flex_attention import create_block_mask, flex_attention
+from src.positional_embeddings import PositionalEmbeddingsTypes, SinusoidalPositionalEncoding, RelPETypeToClass, \
```
andrewargatkiny: It seems that these imports are not required in this file.

andrewargatkiny (changes requested):

src/other_models/attention_kernels.py
```diff
@@ -93,17 +93,21 @@ def __init__(self, config, eps=1e-6):
...
+                attn_mask: torch.Tensor, dropout_p: float, causal: bool, rope_cache):
         # TODO: implement causal linear attention
         queries = self.feature_map(queries)
         queries = nn.functional.dropout(queries,p=dropout_p)
         keys = self.feature_map(keys)
         keys = nn.functional.dropout(keys,p=dropout_p)
         shape = queries.shape
         n, d = shape[-2], shape[-1]
+        if self.apply_relpe_after:
+          queries = rope_cache.apply_relpe(queries)
```
andrewargatkiny: This would apply global RelPE even in case when the attention kernel is invoked from a local layer. At first, please just fix this and send a PR, so we can use it. 

But then, please think of the best way (in terms of speed) to fix this. There are 3 opportunities:
1) Branching (if-else) 
2) A call to a pre-defined wrapper method instead of branching (see how it's done on multiple occasions in `dense_attention.py`)
3) like previous, but unbound function instead of a method.

Could you come up with a way to benchmark these methods and form a conclusion are methods (2) or (3) worth it?

andrewargatkiny (changes requested):

src/other_models/attention_kernels.py
```diff
@@ -38,10 +38,13 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
 class SoftmaxAttention(nn.Module):
     def __init__(self, config):
         super(SoftmaxAttention, self).__init__()
+        self.local = False
```
andrewargatkiny: As we discussed, there's no need for this feature (lines 41-43) for `SoftmaxAttention` and `SlidingWindowAttention`

src/other_models/attention_kernels.py
```diff
@@ -93,21 +100,33 @@ def __init__(self, config, eps=1e-6):
...
+            queries = rope_cache.apply_relpe(queries)
+            keys = rope_cache.apply_relpe(keys)
+          else:
+            queries = rope_cache.apply_relpe2(queries)
+            keys = rope_cache.apply_relpe2(keys)
         if n < d:
             return self.forward_quadratic(queries, keys, values, attn_mask, dropout_p)
         else:
             return self.forward_linear(queries, keys, values, attn_mask, dropout_p)
+    def set_local_relpe_state(self, use_local=True):
```
andrewargatkiny, line 122: Per Python [PEP-8,](https://peps.python.org/pep-0008/#blank-lines) "Method definitions inside a class are surrounded by a single blank line".

src/other_models/modeling.py
```diff
@@ -445,6 +455,8 @@ def __init__(self, config: TransformerConfig, layer_number=1):
         else:
             self.left_pad = self.window_size // 2
             self.right_pad = self.window_size // 2
+        self.attention_kernel.local = True
```
andrewargatkiny: This line is redundant (line 375 in parent class does the job) and doesn't use the proper interface (`set_local_relpe_state` method).

src/other_models/attention_kernels.py
```diff
@@ -93,21 +100,33 @@ def __init__(self, config, eps=1e-6):
         if self.no_reweight:
             self.forward_linear = self._forward_linear_no_norm
             self.forward_quadratic = self._forward_quadratic_no_norm
+        self.apply_relpe_after = config.apply_relpe_after
+        self.local = False
 
     def forward(self, queries: torch.Tensor,
                 keys: torch.Tensor, values: torch.Tensor,
-                attn_mask: torch.Tensor, dropout_p: float, causal: bool):
+                attn_mask: torch.Tensor, dropout_p: float, causal: bool, rope_cache):
```
andrewargatkiny: Please include type hint for `rope_cache` (it's `RelPEBase`).

andrewargatkiny (changes requested):

src/other_models/attention_kernels.py
Rozhkin-Aleksandr: Fixed.

src/other_models/attention_kernels.py
Rozhkin-Aleksandr: Okay, fixed.

src/other_models/modeling.py
Rozhkin-Aleksandr: Fixed.

src/other_models/attention_kernels.py
Rozhkin-Aleksandr: Done.

andrewargatkiny (approved):

PR #21, opened 2025-08-02
Add power, symmetric power, and Taylor_exp-based cores for linear attention

### LINKED ISSUE #27: Incorporate Based and Power Transformer transforms for Linear Attention
We need to add promising recent linear attention transforms which are motivated by calculation of power kernel instead of exponential in attention:

1) Based (https://arxiv.org/abs/2402.18668)
2) Power Transformers / Symmetric Power Transformers (https://manifestai.com/articles/symmetric-power-transformers/ and https://arxiv.org/abs/2507.04239) / PolySketchFormer (https://arxiv.org/abs/2310.01655)

They should be implemented as new transform functions in `src/other_models/attention_kernels.py` file. These transforms should be coded in plain PyTorch.

Power Transformers, Symmetric Power Transformers and basic (exact) variant of PolySketchFormer essentially calculate the same transform. SymPow Transformer tries to bring computational costs down by combining like terms in expansion, but has a more complex algorithm and and is likely slower when implemented in pure PyTorch rather than Triton/Cuda. Therefore, we need to implement Power Transformer at first and proceed with Symmetric Power Transformers later.

The papers from 2) test the architectures only with even powers, so, while implementing general transform for arbitrary power `p`, please instantiate particular transforms in `Transform2Func` register for `p=2` and `p=4`.

mirrtl:


src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+        return torch.cat([term1, term2, term3], dim=-1)
+
+    
+Transform2Func = {
+    None: lambda x: x,
+    "identity": lambda  x: x,
+    "elu": nn.functional.elu,
+    "squared_relu": lambda x: nn.functional.relu(x) ** 2,
+    "1_plus_elu": lambda x: 1 + nn.functional.elu(x),
+    "sym_power_2": SymmetricPowerEmbedding(p=2),
```
andrewargatkiny: When these transforms (lines 122-126) are declared as stateful class instances, as opposed to stateless pure functions, it could lead to all kinds of malevolent side effects in the feature. That's because the class inherits `nn.Module`, and the same instance gets reused in many layers. 

I suggest that it's either reimplemented as pure functions (perhaps, with factory and/ or decorators), or an instance of a corresponding class gets initialized anew for each distinct layer. Please note that all transforms in `Transform2Func` should be either functions or class instances. I think functional form is preferable.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+        self.p = p
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        expanded_x = x
+        
+        for _ in range(self.p - 1):
+            x=x.unsqueeze(-2)
+            expanded_x = expanded_x.unsqueeze(-1) * x
+            
+        norm_factor = math.pow(x.shape[-1], self.p / 2.0)
```
andrewargatkiny: What's the reference for this normalizing factor? I can't find it neither [in the blog](https://manifestai.com/articles/symmetric-power-transformers/) nor in the papers (https://arxiv.org/abs/2507.04239 and https://arxiv.org/abs/2503.03269)

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+class TensorPowerEmbedding(nn.Module):
+    def __init__(self, p: int):
+        super().__init__()
+        self.p = p
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        expanded_x = x
+        
+        for _ in range(self.p - 1):
+            x=x.unsqueeze(-2)
```
andrewargatkiny: Please add spaces around equality sign, per Python's [PEP8](https://peps.python.org/pep-0008/#other-recommendations) style guide.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
     128: sliding_window_mask_128, 256: sliding_window_mask_256,
     512: sliding_window_mask_512, 1024: sliding_window_mask_1024
 }
+
+
+class SymmetricPowerEmbedding(nn.Module):
```
andrewargatkiny, line 36: Add doc-strings for this and other new transforms. And please never forget to properly cite the papers or blogs which introduced the models that you're implementing.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+
+
+class TaylorExp(nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        d = x.shape[-1]
+        
+        r2 = math.sqrt(2)
```
andrewargatkiny: Parameters in lines 103-105 should be pre-calculated only once outside of the `forward` method for efficiency. Please think how to integrate `ModelConfig`s in initialization of these transforms in order to capture `d` value.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+        for _ in range(self.p - 1):
+            x=x.unsqueeze(-2)
+            expanded_x = expanded_x.unsqueeze(-1) * x
+            
+        norm_factor = math.pow(x.shape[-1], self.p / 2.0) 
+
+        return expanded_x.flatten(start_dim=-self.p) / norm_factor
+
+
+class TaylorExp(nn.Module):
```
andrewargatkiny: While this class is called `TaylorExp` in the reference listing of the paper https://arxiv.org/abs/2402.18668, it would be more suitable to call it `Based`.  TaylorExp in general considers any power, and Based (this implementation) considers items only up to second power. 

Please also adjust line 126.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+        dtype = x.dtype
+        
+        # generates list of non-decreasing multiindices
+        indices = list(combinations_with_replacement(range(d), self.p))
+        self.indices_tensor = torch.tensor(indices, device=device, dtype=torch.long)
+        
+        # given a multiindex, counts how many times each index appears
+        counts = torch.zeros(len(indices), d, device=device, dtype=dtype)
+        ones = torch.ones_like(self.indices_tensor, dtype=dtype)
+        counts.scatter_add_(1, self.indices_tensor, ones)
```
andrewargatkiny: This is a very neat implementation! However, it's hard to understand, so I'd add explicit parameter names for `scatter_add_` (dim=1, index=self.indices_tensor, src=ones), and would write a comment detailing what this line of code does, so an unprepared reader could grasp what's happening.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+        
+        # computes multinomial coefficient
+        log_numerator = math.lgamma(self.p + 1)
```
andrewargatkiny, line 82: Please add a comment that `Г(n+1)=n!`. It's also likely that you've used `lgamma` because PyTorch doesn't have native factorial implementation, if that's the case, please also indicate it.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+
+class SymmetricPowerEmbedding(nn.Module):
+    def __init__(self, p: int):
+        super().__init__()
+        self.p = p
+        self.register_buffer('indices_tensor', None)
+        self.register_buffer('coeffs', None)
+        self._initialized = False
+
+    def _initialize_buffers(self, x: torch.Tensor):
```
andrewargatkiny: You can do it in `__init__` without supplying a tensor `x`. You can take `d` from config, and `dtype` should be always `torch.float32`. Since `coeffs` is a PyTorch buffer, when you invoke a command `model.to(...)`, it automatically will cast this tensor to a required format/ device.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
     128: sliding_window_mask_128, 256: sliding_window_mask_256,
     512: sliding_window_mask_512, 1024: sliding_window_mask_1024
 }
+
+
+class SymmetricPowerEmbedding(nn.Module):
+    def __init__(self, p: int):
+        super().__init__()
+        self.p = p
+        self.register_buffer('indices_tensor', None)
```
andrewargatkiny: Use `register_buffer` with option `persistent=False`. It prevents all sorts of errors when saving and then reusing models.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+        self.coeffs = torch.exp(log_coeffs)
+
+        self._initialized = True
+
+    def forward(self, x: torch.Tensor):
+        if not self._initialized:
+            self._initialize_buffers(x)
+        
+        d = x.shape[-1]
+        norm_factor = math.pow(d, self.p / 2.0)
```
andrewargatkiny: See comments for lines 91 and 103.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+        self._initialized = False
+
+    def _initialize_buffers(self, x: torch.Tensor):
+        d = x.shape[-1]
+        device = x.device
+        dtype = x.dtype
+        
+        # generates list of non-decreasing multiindices
+        indices = list(combinations_with_replacement(range(d), self.p))
+        self.indices_tensor = torch.tensor(indices, device=device, dtype=torch.long)
```
andrewargatkiny: Please characterize the shapes of the main tensors and intermediate results of this calculation in comments, similarly to how it's done in `src/dense_attention`.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+
+    def forward(self, x: torch.Tensor):
+        if not self._initialized:
+            self._initialize_buffers(x)
+        
+        d = x.shape[-1]
+        norm_factor = math.pow(d, self.p / 2.0)
+        
+        # Computes the product over the last dimension (monomials of degree p)
+        selected = x[..., self.indices_tensor]
```
andrewargatkiny, line 94: What's the shape of this tensor? Please make a comment.

And for the next line, too.

src/other_models/attention_kernels.py
```diff
@@ -35,6 +30,103 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
     128: sliding_window_mask_128, 256: sliding_window_mask_256,
     512: sliding_window_mask_512, 1024: sliding_window_mask_1024
 }
+
```
andrewargatkiny, line 34: Have you tested that `SymmetricPowerEmbedding` produces the same or  numerically close outputs to `TensorPowerEmbedding`? If yes, for which powers `p`?

src/other_models/attention_kernels.py
andrewargatkiny: On second thought, it seems that only classes inheriting `nn.Module` could accommodate all functionality for more complicated cases, such as `SymmetricPowerEmbedding`. Each layer should initialize its own instance. 

For simple transforms like `identity` or `1_plus_elu`, I'd like to keep a simple function, because wrapping it into a full `nn.Module` causes overhead. Thus, you should create some logic for `self.feature_map`, which checks whether the supplied transform is a function or a class, and in case of a class, instantiates it.

andrewargatkiny (changes requested):

src/other_models/attention_kernels.py
```diff
@@ -77,56 +77,52 @@ def forward(self, x: torch.Tensor):
 
 
 class TensorPowerEmbedding(nn.Module):
-    def __init__(self, p: int):
+    def __init__(self, p: int, config):
         super().__init__()
         self.p = p
+        self.norm_factor = math.pow(config.hidden_size, self.p / 2.0)
```
andrewargatkiny: Here, and everywhere you use `config.hidden_size`: we need a head size, rather than the full hidden size.

src/other_models/attention_kernels.py
```diff
@@ -77,56 +77,52 @@ def forward(self, x: torch.Tensor):
...
-    "identity": lambda  x: x,
-    "elu": nn.functional.elu,
-    "squared_relu": lambda x: nn.functional.relu(x) ** 2,
-    "1_plus_elu": lambda x: 1 + nn.functional.elu(x),
-    "sym_power_2": SymmetricPowerEmbedding(p=2),
-    "sym_power_4": SymmetricPowerEmbedding(p=4),
-    "power_2": TensorPowerEmbedding(p=2),
-    "power_4": TensorPowerEmbedding(p=4),
-    "taylor_exp": TaylorExp()
+    "identity": lambda config=None: lambda x: x,
```
andrewargatkiny: Do we need `None` default value for `config` in all functions, since its only invocation in row 184 always supplies a config?

src/other_models/attention_kernels.py
mirrtl: Done — instances are now initialized anew for each layer.

src/other_models/attention_kernels.py
mirrtl: The coefficient was chosen during testing and has been removed for now.

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done – class constructors now accept a config

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Yes, I have tested it. Experiments were run for p=2. I also verified it with my own tests, which it passed successfully.

src/other_models/attention_kernels.py
mirrtl: Fixed

src/other_models/attention_kernels.py
mirrtl: No, we don’t need a None default for config, fixed.

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
```diff
@@ -36,6 +31,124 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+        
+        self.register_buffer('indices_tensor', None, persistent=False)
+        self.register_buffer('coeffs', None, persistent=False)
+
+        # generates list of non-decreasing multiindices
+        # num_monomials = C(d+p-1, p)
+        # indices: num_monomials, p
+        indices = list(combinations_with_replacement(range(self.d), self.p))
+        indices_tensor = torch.tensor(indices, dtype=torch.long)
+        self.indices_tensor = indices_tensor  # сохраняем в буфер
```
andrewargatkiny: 1) It's generally encouraged to always use only English commentaries in open-source projects targeting global audience, so please rewrite in English.

2) Can you merge lines 66-67?

src/other_models/attention_kernels.py
```diff
@@ -36,6 +31,124 @@ def sliding_window_mask_1024(b, h, q_idx, kv_idx):
...
+
+        # given a multiindex, counts how many times each index appears
+        counts = torch.zeros(len(indices), self.d, dtype=torch.float32)
+        ones = torch.ones_like(indices_tensor, dtype=torch.float32)
```
andrewargatkiny: Please also comment the shape of `ones`. Place it after line 72.

src/other_models/attention_kernels.py
andrewargatkiny: 1) Please make the description (lines before "Args") more concise and explanatory, like "A transform which computes tensor power of degree `p` of an input vector, combining like terms in the expansion. Output vector has C(d+p-1, p) terms."

2) "...and other new transforms" as I've said in the previous comment. Their descriptive part should also be concise and self-explanatory.

andrewargatkiny (changes requested):

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done

src/other_models/attention_kernels.py
mirrtl: Done

andrewargatkiny (approved):

PR #24, opened 2025-08-10
Add value for apply_relpe_after to ensure correct hybrid model operation

### LINKED ISSUE #22: Add option to apply RelPE after feature maps for Linear Transformers

mirrtl:


andrewargatkiny (approved):

andrewargatkiny:
This PR fixes a bug caused by untested case of #22

PR #30, opened 2025-09-03
feat: local_scheme for BertEncoder

glaucomaa:


src/other_models/modeling.py
```diff
@@ -615,7 +617,20 @@ def __init__(self, config):
...
             [copy.deepcopy(layer) for _ in range(config.num_hidden_layers)])
 
-        if config.local_attention:
+        # logic for local attention scheme
+        if hasattr(config, 'local_scheme') and config.local_scheme:
+            scheme = config.local_scheme.split('_')
+            code_map = {
+                'g': BertSelfAttention,
+                'l': BertSelfLocalAttention,
+                'sl': BertSelfShiftedLocalAttention
```
andrewargatkiny: Please add and correctly handle the option for "swa". Keep in mind that it's implemented not as a layer but as an attention kernel.

src/other_models/modeling.py
```diff
@@ -104,6 +104,7 @@ def __init__(self,
                  local_attention=False,
                  window_size=1024,
                  apply_relpe_after=False,
+                 local_scheme=None,
```
andrewargatkiny, line 111: Please add description of this parameter. Make it in similar fashion to the one in `ModelConfig` used for DANet.

src/other_models/modeling.py
```diff
@@ -615,7 +617,20 @@ def __init__(self, config):
...
+        if hasattr(config, 'local_scheme') and config.local_scheme:
+            scheme = config.local_scheme.split('_')
+            code_map = {
+                'g': BertSelfAttention,
+                'l': BertSelfLocalAttention,
+                'sl': BertSelfShiftedLocalAttention
+            }
+            for i, layer in enumerate(self.layer):
+                code = scheme[i % len(scheme)]
+                if code in code_map:
```
andrewargatkiny: Better to drop this line so if user makes a mistake in a config, it won't go on silently.

andrewargatkiny (approved):

andrewargatkiny (changes requested):
.

andrewargatkiny (approved):

PR #33, opened 2025-09-13
Weight norms logging enhancement

### LINKED ISSUE #29: Weight norm logging improvement for models other than DANet.
1) Lately, weight norm logging places graphs for all parameters, even for embeddings and layers on the same figure, even for DANet. Please investigate and resolve this bug.

Image

2) Before that, weight norm logging worked best for DANet, wherein it correctly grouped similar type parameters for each layer:

Image

However, for other architectures, including vanilla and linear Transformers, it places the graphs for all possible parameters and values on one figure, which renders their exploration and analysis incomprehensible. Especially, if a layer has many distinct parameters (greater quantity than DANet), it would be beneficial to specify, which of layers' parameters should form a distinct figure. It can be specified e.g., in an optional sub-dictionary of `training` part of a model config. 

For example, we could like that all LayerNorms in all layers of a Transformer model should go to one figure, and all parameters of FFNs – to another, and all constituents of the attention sub-block – to a third one.

Image

herbitovich:
Fixes #29

### DIFF
--- a/deepspeed_train.py
+++ b/deepspeed_train.py
@@ -526,29 +526,21 @@ def report_model_weights(args, model, step, bins=20):
         layers = attrgetter(backbone.PATH_TO_LAYERS)(backbone)
         embeddings = attrgetter(backbone.PATH_TO_EMBEDDINGS)(backbone)
 
-        """
-        Get the full attribute path (in terms of the `model` object) for a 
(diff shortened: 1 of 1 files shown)


andrewargatkiny (approved):
Nice work!

Issue #35, opened 2025-09-30
Refactor of high-level model classes to accommodate to different layer types.


andrewargatkiny:
Difficulties with clean integration of S4 layers highlight the necessity to have a single interface model which could use different types of sequence-mixer layers (Transformers, Linear Attention variants, SSMs, DANet, etc.). Right now we have nearly duplicate classes for DANet, Transformer-based and HuggingFace models (e.g., `DANetForPreTraining`, `TransformerForPreTraining`, `HFForPreTraining`), and a multitude  of largely the same tasks at `utils/tasks.py`. 

If we were to implement an independent backbone, mlm, clm, classification, regression, etc. classes for each new model type, it would be overwhelmingly redundant and render incorporation of new functionality (e.g. causal inference, which I'm now working on) troublesome and error-prone.

Instead, I'm proposing to refactor existing classes in `src/other_models/modeling.py`, such as  `TransformerConfig`, `BertEncoder` and `TransformerForPreTraining`, so they can work with arbitrary sequence-mixer layers. Here's the tentative plan:

1) If a model config doesn't have doesn't have the key `layers`, it falls back to legacy implementation.

2) If the key is present, than it contains a list of individual layer configs (Python/ json dicts). Each layer config should contain all information required for an instantiation of the specified layer type. The layer type should be specified by a `layer_type` key.  There should be an individual layer config class for each of the layer types. GOTCHA: some layer types might need whole-model-level information for instantiation, for example DANet and Transformer layers use number of layers for initialization of some weights.

3) The layer config should also contain a `layer_name` field. The only exemption is possible if a model uses a single layer type. Otherwise, this field would be used in `layers_scheme`  (which should have similar purpose and replace the legacy `local_scheme` in this new implementation). That way, we can declare any type of hybrid model (e.g., the one which simultaneously includes DANet, Transformer, and SSM layers) , or homogenous-type model with different hyper-parameters (e.g., number of heads, type of locality and window size) for different layers.

4) There should be a standalone file with the registry for all of the existing layer types and layer config types.

5) Many layers require specific initialization. Hence, `init_bert_weights` method should not interfere with individual layers. N.B.: for Transformer-based layers, some parameters inside the layer are currently initialized with exactly this function, so it makes sense to duplicate and call the relevant initialization code *inside* affected parts of `BertLayer` and to not use the init function on the outer,`BertEncoder` layer. 

[...]

Issue #41, opened 2025-10-06
Fix a few bugs


andrewargatkiny:
1. Method `_forward_chunked` in `src/dense_attention.py` does not work correctly, when supplied with sequence length not divisible by chunk_size/ window_size – it throws an exeption.
2. In case of several heads in DenseAttention, RoPE get applied wrong way: for an input two-dimensional vector (x1, x2) the two components of output vector $(x_1 \cos \theta - x_2 \sin \theta, x_2 \cos \theta + x_1 \sin \theta)$ end up in different heads. This bug is specific to DenseAttention only.
3. Saving checkpoints in `deepspeed_train.py` is implemented by infinite loop with try-except block, which is helpful during a sporadic system glitch, but leads to excessive system calls and enormous log files if the cause of error is lack of space on the hard drive. Some sleep interval needs to be added to mitigate it.

andrewargatkiny:
All done. Additionally, I 
* made an initialization of `RoPE` and Trigonometric RelPE functions more intuitive and no longer exposing inner logic (previously, if you needed RoPE to work with inputs having separate head and embedding dims, you'd have to supply `num_heads=None` which is confusing);
* modified `apply_local_relpe` method so it can work with both separate and merged `num_heads` and  `embed_dim` dimensions;
* created tests for RoPE in all available configurations.

See #45 for full changelog.

Tests can be launched with `PYTHONPATH=$PWD pytest -v tests`

I made tests for only `RoPE` class of relative positional embeddings. The function `test_rope_all_methods_one_func` should be augmented and renamed so it can additionally test `CosRelPE` and `CosMinusSinRelPE`. Reference class for Trigonometric RelPEs should be taken from the last commit in `dev` branch before #45 was merged into it.

PR #42, opened 2025-10-06
Refactoring modeling.py for new layers initialization mechanism

Rozhkin-Aleksandr:
Solving the #35 issue.
I have changed pipeline in `modeling.py` to support new config arguments: layers and layers_scheme.
I have changed the initialization mechanism, I've taken it out `BertEncoder`.
The legacy realization can be found in `modeling_legacy.py`.
The `layers_registry.py` was created.
New class-wrapper `DANetLayerForMixing` was added into `danet_layers.py` to incorporate DANet layers with local attention in new pipeline with backward compatibility.


src/other_models/modeling.py
```diff
@@ -109,8 +111,10 @@ def __init__(self,
                  window_size=1024,
                  apply_relpe_after=False,
                  local_scheme=None,
+                 layers_scheme=None,
                  power=2,
                  scaling_d_factor=False,
+                 layers=None,
```
andrewargatkiny, line 100: Please document the new parameters in multi-line comment below where all other params are explained.

src/other_models/layers_registry.py
```diff
@@ -0,0 +1,14 @@
+from ..danet_layers import DANetLayerForMixing
+from ..model_config import DANetLayerConfig
+from ..other_models.modeling import BertLayer, TransformerLayerConfig
+
+
+LAYER_TYPE2CLASS = {
```
andrewargatkiny: Please rename it to `LayerTypeToClass` and `LayerConfigToClass` for consistency with other similar registries' names in our framework.

src/other_models/modeling.py
```diff
@@ -233,6 +239,142 @@ def to_json_string(self):
         """Serializes this instance to a JSON string."""
         return json.dumps(self.to_dict(), indent=2, sort_keys=True) + "\n"
 
+class TransformerLayerConfig(object):
```
andrewargatkiny: Everything related to Transformer layers (config, the whole layer class itself, and component layers) is better to move to a new dedicated file, e.g. `transformers.py`. The current file at this point became really huge.

src/other_models/modeling.py
```diff
@@ -233,6 +239,142 @@ def to_json_string(self):
...
+            max_position_embeddings: The maximum sequence length that this model might
+                ever be used with. Typically set this to something large just in case
+                (e.g., 512 or 1024 or 2048).
+            type_vocab_size: The vocabulary size of the `token_type_ids` passed into
+                `BertModel`.
+            initializer_range: The sttdev of the truncated_normal_initializer for
+                initializing all weight matrices.
+            apply_relpe_after: Whether Relative Positional Encoding (RELPE) is applied after the feature map (if true)
+             or before the linear attention kernel (if false).
+            local_scheme: Scheme to form patterns of local and global attention
```
andrewargatkiny: There's no `local_scheme` parameter in this config. Please also check carefully this doc and find if there are other deprecated descriptions. The set of descriptions should fully conform to the set of init arguments.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
-        # fallback to old logic for backward compatibility
-        elif config.local_attention:
-            for i, layer in enumerate(self.layer):
-                if i % 3 == 0:
-                    layer.attention.self = BertSelfLocalAttention(config)
-                elif i % 3 == 1:
-                    layer.attention.self = BertSelfShiftedLocalAttention(config)
-
+            self.rope_caches = []
+            # Lazy import to avoid circular import issues
```
andrewargatkiny: This is producing circular import issues because the classes for config and the layer(s) are also in this file, and they are imported in `layers_registry`. After you've relocated these classes to a separate file, please move this import statement to the top of this file, where all other imports are situated. It's always highly recommended to have all imports in one place for ease of understanding the scope of dependencies.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
 
         #Added later to make it similar to GPT-2
         self.FinalLayerNorm = BertLayerNorm(config.hidden_size, eps=1e-12)
-        if (config.pos_emb_type == PositionalEmbeddingsTypes.RELPE and
-                config.relpe_type is not None):
-            self.relpe_type = RelPEType[config.relpe_type.upper()]
+        
+
+
+        if config.layers is None:
```
andrewargatkiny: Just for clarity, please swap the statements (`if config.layers is not None`), so that the default case is the new logic, and the legacy one is left in `else` branch.

andrewargatkiny: Please also comment at the start of that branch that it's the old logic left for compatibility with existing configs.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
-
+            self.rope_caches = []
+            # Lazy import to avoid circular import issues
+            from .layers_registry import LAYER_TYPE2CLASS, LAYER_TYPE2CONFIG_CLASS
+
+            name_to_config = {}
+            for layer in config.layers:
+                if "layer_type" not in layer:
+                    raise ValueError(f"Layer '{layer['layer_name']}' must contain 'layer_type'")
+                layer_name = layer["layer_name"]
```
andrewargatkiny: If there's only one layer type in `config.layers` and  `layers_scheme` is null or empty, we assume that all layers in the model being initialized are of the same type and proceed instead of casting an exception. For that case, there should be a default option (e.g., `None`) if no value is supplied in config for `layer_name`.

However, if `layers_scheme` has some value(s), then all layer types should have non-null `layer_name` values.

Please note that more layer types may be declared in a config, then number of names in `layers_scheme`. For example, we can declare all available types of layers at once for convenience, and then test in different combinations, just changing `layers_scheme` at launch time.

It's also possible that number of layer types could be less than then number of names in `layers_scheme` if the names are not unique (e. g., "sw_sw_g").

src/other_models/modeling.py
```diff
@@ -233,6 +239,142 @@ def to_json_string(self):
...
+        self.pre_attn_ln_type = pre_attn_ln_type
+        self.post_attn_ln_type = post_attn_ln_type
+        self.causal = causal
+        self.local_attention = local_attention
+        self.window_size = window_size
+        self.apply_relpe_after = apply_relpe_after
+        self.power = power
+        self.scaling_d_factor = scaling_d_factor
+        
+    @classmethod
```
andrewargatkiny: Turns out that the methods on lines 349-377 are not required, because the instance is constructed directly with __init__ method in the code. Please remove these spare methods for now.

src/model_config.py
```diff
@@ -233,3 +233,189 @@ def to_dict(self):
...
+        self.local_attention = local_attention
+        self.window_size = window_size
+        self.local_relpe = local_relpe
+        self.local_scheme = local_scheme
+        self.hybrid = hybrid
+        self.transformer_heads = transformer_heads
+        self.attn_proj_biases = attn_proj_biases
+        self.locality = locality
+
+    @classmethod
```
andrewargatkiny: Turns out that the methods on lines 402-421 are not required, because the instance is constructed directly with init method in the code. Please remove these spare methods for now.

src/other_models/modeling.py
```diff
@@ -625,9 +767,32 @@ def __init__(self, config):
             self.intermediate = BertSwigluUp(config)
             self.output = BertSwigluDown(config)
 
-    def forward(self, hidden_states, attention_mask, rope_cache):
+        # Local, layer-specific initialization: duplicate the relevant parts
```
andrewargatkiny: This comment is hard to interpret without knowing the context about past methods of initialization, thus it's better to remove it.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
+                    raise ValueError(f"Layer '{layer['layer_name']}' must contain 'layer_type'")
+                layer_name = layer["layer_name"]
+                if layer_name in name_to_config:
+                    raise ValueError(f"Duplicate layer_name detected in config.layers: '{layer_name}'")
+                name_to_config[layer_name] = layer
+
+            ordered_names = config.layers_scheme.split("_")
+            if len(ordered_names) == 0:
+                raise ValueError("layers_scheme must be a non-empty string of layer_name identifiers")
+            repeats = config.num_hidden_layers // len(ordered_names)
```
andrewargatkiny: This routine should also account for remainder layers if the total number of layers is not divisible by `layers_scheme` cycle length. Currently, they are dismissed, e.g., if `num_hidden_layers=32`, and `len(ordered_names)=3`,  the model won't have layers 30 and 31.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
+            modules = []
+            for _ in range(repeats):
+                for layer_name in ordered_names:
+                    layer_config_dict = name_to_config[layer_name]
+                    layer_type = layer_config_dict["layer_type"]
+                    model_config_dict = config.to_dict()
+                    model_config_dict.update(layer_config_dict)
```
andrewargatkiny: This could be problematic due to several reasons:
1. (Major concern) If some attribute names in layer and model configs coincide, the value from the layer config would silently override its counterpart from the model config, which is unintended behavior. 
2. (Minor concern) `model_config_dict` is constructed by copying the `__dict__` of the model class instance. Basically, it's an entire representation of the class and instance, which contains not only visible params, but all sorts of dunder variables (e.g., `__init__`) and all of the methods and objects of the original class. Having all these unnecessary params in the model config is potentially error-prone and harder to debug.

Instead, I suggest that you parse the signature of the init method of relevant layer class, check which its parameters were not explicitly declared in the layer config dict, and then search for these parameters in the outer layer config.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
+                    config_class = LAYER_TYPE2CONFIG_CLASS[layer_type]
+                    layer_config = config_class(**model_config_dict)
+
+                    if (model_config_dict["pos_emb_type"] == PositionalEmbeddingsTypes.RELPE and
+                            model_config_dict["relpe_type"] is not None):
+                        relpe_type = RelPEType[model_config_dict["relpe_type"].upper()]
+                    else:
+                        relpe_type = RelPEType.DUMMY
+                    if model_config_dict["layer_type"] in ["danet", "danet_with_local_attention"]:
```
andrewargatkiny: There's no "danet_with_local_attention" layer type in the registry. So I guess this could be just "danet".

andrewargatkiny: Unrelated comment: since the code `RelPETypeToClass[relpe_type]` gets repeated two times, let's make it more readable and clear: before this conditional statement you may write `relpe_class = RelPETypeToClass[relpe_type]` and then use `relpe_class`.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
-                # 'g' is the default and requires no change, so we just pass.
-        # fallback to old logic for backward compatibility
-        elif config.local_attention:
-            for i, layer in enumerate(self.layer):
-                if i % 3 == 0:
-                    layer.attention.self = BertSelfLocalAttention(config)
-                elif i % 3 == 1:
-                    layer.attention.self = BertSelfShiftedLocalAttention(config)
-
+            self.rope_caches = []
```
andrewargatkiny: Unfortunately, plain Python lists of `nn.Module`s cannot be bound properly to an outer `nn.Module` and preserve all functionality of PyTorch (e.g., to cast list member modules to specific datatype or move to a device along with the outer module). For example, if you try to use a configuration with RoPE embeddings for training on GPU, it will throw an error because RoPE cache will stay on CPU, while all other parameters will move to GPU.

So please at first accumulate the individual rope_caches into a list, then wrap it into `nn.ModuleList`, and finally assign this ModuleList to the `self.rope_caches`.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
+                            model_config_dict["hidden_size"] // model_config_dict["num_attention_heads"],
+                            num_heads=model_config_dict["num_attention_heads"]
+                        )
+                    else:
+                        rope_cache = RelPETypeToClass[relpe_type](
+                            model_config_dict["max_position_embeddings"], #args.max_seq_length
+                            model_config_dict["hidden_size"] // model_config_dict["num_attention_heads"],
+                            num_heads=None
+                        )
+                    self.rope_caches.append(rope_cache)
```
andrewargatkiny: In this implementation, the number of RoPE modules is equal to the number of layers. But the number of unique modules among them is equal to the number of layer types at most, which leads to less efficient use of memory. 

It would be more efficient to initialize only the unique RoPE modules and then reuse them.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
+            
+            self.layer = nn.ModuleList(modules)
+            # Save window_size for mask preparation in forward
+            self.window_size = getattr(config, 'window_size', 1024)
+            
     def forward(self,
                 hidden_states,
                 attention_mask,
-                output_all_encoded_layers=True):
+                output_all_encoded_layers=True, **kwargs):
```
andrewargatkiny: Please put `**kwargs` on a separate line for style consistency.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
+                        )
+                    self.rope_caches.append(rope_cache)
+
+                    layer_class = LAYER_TYPE2CLASS[layer_type]
+                    module = layer_class(layer_config)
+                    modules.append(module)
+            
+            self.layer = nn.ModuleList(modules)
+            # Save window_size for mask preparation in forward
+            self.window_size = getattr(config, 'window_size', 1024)
```
andrewargatkiny: `self.window_size` does not gets used anywhere further in the code. 

But it reveals a major problem: mask formation is different in the DANet and Transformer-like models. Current masking procedure in this file is intended only for Transformers. Please study what principles govern the formation and usage of masks in DANet and pay special attention to differences between masking for local and global layers.

Ideally, the number of masks should coincide with the number of layer types, and each layer type's `forward` should be ingested with its own type of mask.

src/other_models/modeling.py
```diff
@@ -638,64 +803,124 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
+                    else:
+                        rope_cache = RelPETypeToClass[relpe_type](
+                            model_config_dict["max_position_embeddings"], #args.max_seq_length
+                            model_config_dict["hidden_size"] // model_config_dict["num_attention_heads"],
+                            num_heads=None
+                        )
+                    self.rope_caches.append(rope_cache)
+
+                    layer_class = LAYER_TYPE2CLASS[layer_type]
+                    module = layer_class(layer_config)
```
andrewargatkiny: In the new workflow, there is no mechanics to choose `local` or `local_shifted` locality type for Transformer-like models. Please fix it. Note that the old implementation was patchy and assigned appropriate attention classes during init of this module, and now it needs to be done properly, like in DANet layers.

andrewargatkiny: Please also think of other use cases that could be overlooked in this implementation in comparison with the previous.

src/other_models/modeling.py
andrewargatkiny: Please change the function name to a more generic (this architecture had transcended bert use case ages ago).

andrewargatkiny: Let's also remove the check in this line and make the statements below not dependable on the distributed rank. While it doesn't change anything under the hood (DeepSpeed and torch's DDP broadcast parameters from rank 0 to other nodes anyway), it would make the code cleaner and less confusing.

src/danet_layers.py
```diff
@@ -64,8 +93,36 @@ def __init__(self, config: ModelConfig, layer_number: int=0):
...
+                module.weight.data.normal_(mean=0.0, std=std)
+                if module.bias is not None:
+                    module.bias.data.zero_()
+            # Initialize layer norms
+            elif isinstance(module, BertLayerNorm):
+                module.bias.data.zero_()
+                module.weight.data.fill_(1.0)
 
     def forward(self, hidden_states, attention_mask, rope_cache=None):
+        if attention_mask.dim() == 4:
```
andrewargatkiny: This method of Transformer's mask adjustment to DANet doesn't account for scaling by $N^{-1/3}$ where N could be length of the whole sequence or window size for particular layer.

src/danet_layers.py
```diff
@@ -1,27 +1,56 @@
...
         self.attention = DenseAttention(config, layer_number=layer_number)
         self.ffn = SwiGLU(config) if config.swiglu_ffn else ExpandedFFN(config)
         self.ffn_activation = Activation2Class[config.post_attn_ln_type](config.hidden_size)
+        
+        # Local, layer-specific initialization: duplicate the relevant parts
+        # from the global initializer, without relying on outer apply().
+        if config.locality is not None:
+            self._init_weights(config)
+    
+    def _init_weights(self, config):
```
andrewargatkiny: This init function is taken from the Transformer `src/other_models/modeling.py` file. DANet has its own dedicated modeling file `src/modeling.py` with the correct weight initialization method.

src/danet_layers.py
```diff
@@ -1,27 +1,56 @@
...
+                module.weight.data.normal_(mean=0.0, std=std)
+                if module.bias is not None:
+                    module.bias.data.zero_()
+            # Initialize layer norms
+            elif isinstance(module, BertLayerNorm):
+                module.bias.data.zero_()
+                module.weight.data.fill_(1.0)
 
     def forward(self, hidden_states, attention_mask, rope_cache=None):
+        if attention_mask.dim() == 4:
```
andrewargatkiny: This method of Transformer's mask adjustment to DANet doesn't account for scaling by $N^{-1/3}$ where N could be length of the whole sequence or window size for particular layer.

src/other_models/modeling.py
```diff
@@ -1387,7 +1621,8 @@ def __init__(self, config, args):
         self.bert = BertModel(config)
         self.dropout = nn.Dropout(config.hidden_dropout_prob)
         self.classifier = nn.Linear(config.hidden_size, self.num_labels)
-        self.apply(self.init_bert_weights)
+        # Initialize only head; encoder layers self-initialize.
+        self.classifier.apply(self.init_bert_weights)
 
     def forward(self,
                 input_ids,
```
andrewargatkiny, line 1165: All `forward`s in the outermost layers (not only `TransformerForPreTraining`) may include can include different `**kwargs`  for different types of layers.

src/danet_layers.py
```diff
@@ -1,27 +1,56 @@
...
-from src.model_config import ModelConfig
+from src.model_config import DANetLayerConfig
 from src.positional_embeddings import RoPE
 
 
 class DANetLayer(nn.Module):
     """Basic DenseAttention Network layer which can be put into a model as a
     replacement to a standard Transformer block."""
-    def __init__(self, config: ModelConfig, layer_number: int=0):
+    def __init__(self, config: DANetLayerConfig, layer_number: int=0):
```
andrewargatkiny: As we discussed, there should be a full backward compatibility, and no changes at all in current `DANet` classes. All features for integration with new workflow should be relegated to the dedicated wrapper class. Specifically, this includes weight initialization function and `ModelConfig` to `DANetLayerConfig` conversion. And I'm not sure if the conversion is required at all. What are the differences between the two config classes?


And, as mentioned above, it would be preferable if the mask formation logic transpire in the `BertEncoder` class.

configs/assets/Matmuls_are_Enough_ICML_Jun2025_ CameraReady.pdf
andrewargatkiny, line 1: What have you changed in this file?

src/danet_layers.py
```diff
@@ -39,7 +68,7 @@ class DANetLayerWithLocalAttention(nn.Module):
         'g': 'global', 'l': 'local', 'sl': 'shifted_local',
         'sw': 'sliding_window', 'softmax': 'softmax'
     }
-    def __init__(self, config: ModelConfig, layer_number: int=0):
+    def __init__(self, config: DANetLayerConfig, layer_number: int=0):
```
andrewargatkiny: See comment for line 18.

src/danet_layers.py
```diff
@@ -76,6 +133,34 @@ def forward(self, hidden_states, attention_mask, rope_cache=None):
...
         return hidden_states
 
+
+class DANetLayerForMixing(nn.Module):
+    """
+    Wrapper over DANetLayer and DANetLayerWithLocalAttention for compatibility with new style of layers modeling.
+    """
+    def __init__(self, config: DANetLayerConfig):
+        super(DANetLayerForMixing, self).__init__()
+        if config.locality == "local":
```
andrewargatkiny: In fact, lines 143-158 can be deleted if we just directly populate `local_scheme` attribute in a DANet layer config with the appropriate values (keys in `code_to_layer` dict) instead of the new `locality` parameter.

And, as documented on line 66, we don't need `DANetLayer` class for global layer type – it works perfectly fine with `DANetLayerWithLocalAttention` class.

Furthermore, please investigate if with the removal of `locality` attribute in `DANetLayerConfig`, what are the differences between this config and the `ModelConfig`? If there are no differences, we can just delete `DANetLayerConfig` and register `ModelConfig` as the layer config for DANet.

configs/lra/hybrid/ds_train_transformer_pathfinder32_layers.sh
andrewargatkiny, line 1: This is based on the old version of the launch script which does not support override functionality and, probably, something else. Please consider using the latest version instead.

andrewargatkiny (changes requested):

src/other_models/modeling.py
Rozhkin-Aleksandr: Added an explanation of these parameters.

src/other_models/layers_registry.py
Rozhkin-Aleksandr: Done.

src/other_models/modeling.py
Rozhkin-Aleksandr: Created new file transformers.py and moved configs and realization of transformer layer into it.

src/other_models/modeling.py
Rozhkin-Aleksandr: Removed this  line and checked conformity of set of descriptions and set of arguments.

src/other_models/modeling.py
Rozhkin-Aleksandr: A file transformers.py is very helpful for solution of circular import issue! I moved import at the top of the file.

src/other_models/modeling.py
Rozhkin-Aleksandr: Done.

src/other_models/modeling.py
Rozhkin-Aleksandr: Added support of two use cases: layers_scheme is not None, and layers_scheme is None + len(layers) == 1.

src/other_models/modeling.py
Rozhkin-Aleksandr: Done.

src/model_config.py
Rozhkin-Aleksandr: ModelConfig class is registered as config class for danet layer, so DanetLayerConfig is no longer in use.

src/other_models/modeling.py
Rozhkin-Aleksandr: Dropped it.

src/other_models/modeling.py
Rozhkin-Aleksandr: Added raising the exception in case, when number of layers in layers scheme is not a divisor of number of hidden layers.

src/other_models/modeling.py
Rozhkin-Aleksandr: Added parsing of signature and fixed this bugs.

src/other_models/modeling.py
Rozhkin-Aleksandr: Fixed.

src/other_models/modeling.py
Rozhkin-Aleksandr: Added using pytorch module lists for rope_caches and layers.

src/other_models/modeling.py
Rozhkin-Aleksandr: Changed the logic to create only one rope_cache for each layer name.

src/other_models/modeling.py
Rozhkin-Aleksandr: Okay, done.

src/other_models/modeling.py
Rozhkin-Aleksandr: Added new logic of creation of masks in the forward method of BertEncoder.

src/other_models/modeling.py
Rozhkin-Aleksandr: Now it's name is init_weights.

Rozhkin-Aleksandr: Done.

src/danet_layers.py
Rozhkin-Aleksandr: Added new logic of masks creation.

src/danet_layers.py
Rozhkin-Aleksandr: Fixed.

src/other_models/modeling.py
Rozhkin-Aleksandr: Fixed.

src/danet_layers.py
Rozhkin-Aleksandr: Dropped the DANetLayerConfig at all.

configs/assets/Matmuls_are_Enough_ICML_Jun2025_ CameraReady.pdf
Rozhkin-Aleksandr: I just accidentally copied it into configs directory, in this commit I fixed it. I have not changed anything in it.

configs/lra/hybrid/ds_train_transformer_pathfinder32_layers.sh
Rozhkin-Aleksandr: Fixed.

src/danet_layers.py
Rozhkin-Aleksandr: Changed the logic for using local_scheme instead of new parameter locality.

src/other_models/modeling.py
Rozhkin-Aleksandr: Fixed. Now we can use local attention in transformer models with the "local_scheme" parameter, just like we do for Danet models.

src/other_models/modeling.py
```diff
@@ -815,11 +447,9 @@ def init_bert_weights(self, module):
             # cf https://github.com/pytorch/pytorch/pull/5617
             num_layers = self.config.num_hidden_layers
             std = self.config.initializer_range
-            if hasattr(module, 'bert_output_layer'):
```
andrewargatkiny, line 818: This condition should have been kept for the legacy case.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                self.repeats = config.num_hidden_layers
+                
+                if (config.get("pos_emb_type") == PositionalEmbeddingsTypes.RELPE and
+                        final_layer_params.get("relpe_type") is not None):
+                    relpe_type = RelPEType[final_layer_params["relpe_type"].upper()]
+                else:
+                    relpe_type = RelPEType.DUMMY
+                relpe_class = RelPETypeToClass[relpe_type]
+                if layer_type == "danet":
+                    rope_cache = relpe_class(
```
andrewargatkiny: This class is not instantiated accordingly to new `__init__` method introduced in #41 .

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                        final_layer_params["max_position_embeddings"],
+                        final_layer_params["hidden_size"] // final_layer_params["num_attention_heads"],
+                        num_heads=final_layer_params["num_attention_heads"]
+                    )
+                else:
+                    rope_cache = relpe_class(
```
andrewargatkiny: Please see comment on line 222.

src/other_models/transformers.py
```diff
@@ -0,0 +1,726 @@
...
+                scores by their row-wise sums.
+            no_reweight_post_norm: In case of enabled `no_reweight` option in linear
+                attentions, determines whether and which layer norm to use at the end
+                of attention kernel computation. Defaults to None.
+            apply_relpe_after: For linear attentions, determines whether Relative
+                Positional Encoding (RELPE) is applied after the feature map (if true)
+                or before the linear attention kernel (if false).
+            layers_scheme: Defines the sequence and repetition of layers within the encoder.
+                This should be a string of layer names separated by underscores
+                (e.g., 'attention_ffn_attention'). Each name must correspond to a unique layer_name
```
andrewargatkiny: This example might be misleading. Currently in the framework, FFN sub-block is treated as a part of a layer (DANet, Transformer, etc.) rather than a standalone layer. The example gives the opposite impression. A better neutral example could be 'layer1_layer2_layer1_layer3', or 'danet_swa32tr_swa32tr'.

src/other_models/transformers.py
```diff
@@ -0,0 +1,726 @@
...
+
+
+def swish(x):
+    return x * torch.sigmoid(x)
+
+
+ACT2FN = {"gelu": gelu, "relu": torch.nn.functional.relu, "swish": swish}
+
+
+class TransformerConfig(object):
```
andrewargatkiny: This class belongs to `src/other_models/modeling.py`.

src/other_models/transformers.py
```diff
@@ -0,0 +1,726 @@
...
+                This should be a string of layer names separated by underscores
+                (e.g., 'attention_ffn_attention'). Each name must correspond to a unique layer_name
+                key in one of the configuration dictionaries provided in the layers parameter.
+                power: For Power Attention, determines the power (p).
+            scaling_d_factor: For Power Attention, determines whether to scale q,k by a
+                predetermined scaling factor depending on d for additional numerical
+                stability.
+            layers: A list of dictionaries, where each dictionary provides the configuration
+                for a specific layer type. Each dictionary must contain a unique layer_name
+                key, which is then used by the layer_scheme parameter to construct the full encoder stack.
```
andrewargatkiny: Our framework adopts a documentation style in which mentions of parameters should be enclosed in backtick (\`) signs, i.e. \`layer_scheme\`.

src/other_models/transformers.py
```diff
@@ -0,0 +1,726 @@
...
+            self.attention_kernel = attention_kernel
+            self.feature_map = feature_map
+            self.no_reweight = no_reweight
+            self.no_reweight_post_norm = no_reweight_post_norm
+            self.embedding_dropout = embedding_dropout
+            self.hidden_dropout_prob = hidden_dropout_prob
+            self.attention_probs_dropout_prob = attention_probs_dropout_prob
+            self.attn_proj_biases = attn_proj_biases
+            self.max_position_embeddings = max_position_embeddings
+            if layers is None:
```
andrewargatkiny: While each of layers could have a separate type of Relative Positional embeddings, it should not influence the `BertEmbeddings` layer which still can have Learned, Sinusoidal or no positional embeddings.

src/other_models/transformers.py
```diff
@@ -0,0 +1,726 @@
...
+                 pre_attn_ln_type="default",
+                 post_attn_ln_type="default",
+                 causal=False,
+                 local_attention=False,
+                 window_size=1024,
+                 apply_relpe_after=False,
+                 max_position_embeddings=512,
+                 pos_emb_type="learned",
+                 relpe_type=None,
+                 local_scheme=None,
```
andrewargatkiny, line 90: There should be a description of this parameter in the docstring.

src/other_models/transformers.py
```diff
@@ -0,0 +1,726 @@
...
+        if config.post_attn_ln_type != "default":
+            self.PostAttentionLayerNorm = Activation2Class[
+                config.post_attn_ln_type](config.hidden_size, eps=1e-12)
+        self.intermediate = BertIntermediate(config)
+        self.output = BertOutput(config)
+        if config.hidden_act == "swiglu":
+            self.intermediate = BertSwigluUp(config)
+            self.output = BertSwigluDown(config)
+        # logic for local attention scheme
+        if hasattr(config, 'local_scheme') and config.local_scheme and len(config.local_scheme.split("_"))==1:
```
andrewargatkiny: As this is a new piece of code, invoked only when `hasattr(config, 'local_scheme')`,  there are no requirements for backward compatibility in this case. Thus, we don't have to assume that input string holds "_"-separated sequences. So, there should be no splits performed on this line and the next one. The whole string should be treated as a single code.

src/danet_layers.py
```diff
@@ -117,3 +117,22 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
 
+
+class DANetLayerWrapper(nn.Module):
+    def __init__(self, config: ModelConfig):
+        super(DANetLayerWrapper, self).__init__()
+        self.config = config
+        self.layer = DANetLayerWithLocalAttention(config)
+        self.apply(self.init_bert_weights)
+
+    def init_bert_weights(self, module):
```
andrewargatkiny: `init_weights` or `init_danet_weights`

src/danet_layers.py
```diff
@@ -117,3 +117,22 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
+        std = self.config.initializer_range
+        if isinstance(module, nn.Embedding):
+            module.weight.data.normal_(mean=0, std=std)
+        elif isinstance(module, nn.Linear):
+            module.weight.data.uniform_(-std, std)
+            if module.bias is not None:
+                module.bias.data.zero_()
+
+    def forward(self, hidden_states, attention_mask, rope_cache=None):
```
andrewargatkiny: There's one nuance you have left unnoticed. Can you spot it?

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                modules.append(module)
+                modules = modules * self.repeats
+                self.rope_caches = nn.ModuleList(rope_caches * self.repeats)
+                self.layer_scheme = [single_layer_config]
+            else:
+                raise ValueError("Invalid layer configuration. Provide either a 'layers_scheme' with corresponding layer definitions, or a single layer definition in 'config.layers' without a 'layers_scheme'.")
+
+            self.layer = nn.ModuleList(modules)
+
         else:
```
andrewargatkiny, line 379: Please add a comment that it's for the case of legacy configs, and specify conditions, under which this branch would be invoked.

src/other_models/modeling.py
andrewargatkiny: It seems you forgot to leave the comment at the start of the branch

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+            rope_caches = []
+            modules = []
+
+            if config.layers_scheme:
+                name_to_config = {}
+                for layer in config.layers:
+                    if not layer.get("layer_name"):
+                        raise ValueError("Each layer in config.layers must have a 'layer_name' when 'layers_scheme' is used.")
+                    if "layer_type" not in layer:
+                        raise ValueError(f"Layer '{layer['layer_name']}' must contain 'layer_type'")
```
andrewargatkiny: ... must contain 'layer_type' attribute value

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                    
+                    layer_name = layer["layer_name"]
+                    if layer_name in name_to_config:
+                        raise ValueError(f"Duplicate layer_name detected in config.layers: '{layer_name}'")
+                    name_to_config[layer_name] = layer
+
+                ordered_names = config.layers_scheme.split("_")
+                
+                if config.num_hidden_layers % len(ordered_names) != 0:
```
andrewargatkiny: As discussed earlier, the model should work and properly use all of the `num_hidden_layers`, even if `num_hidden_layers % len(ordered_names) != 0`. An example: `num_hidden_layers = 5, layers_scheme = "l_sl_g"`, the model still should use all 5 layers. The unrolled scheme would be "l_sl_g_l_sl".

src/other_models/modeling.py
andrewargatkiny: Unfortunately, that would not be the intended behavior. Please see new comments.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                    if (final_layer_params.get("pos_emb_type") == PositionalEmbeddingsTypes.RELPE and
+                            final_layer_params.get("relpe_type") is not None):
+                        relpe_type = RelPEType[final_layer_params["relpe_type"].upper()]
+                    else:
+                        relpe_type = RelPEType.DUMMY
+
+                    relpe_class = RelPETypeToClass[relpe_type]
+                    if layer_type == "danet":
+                        rope_cache = relpe_class(
+                            final_layer_params["max_position_embeddings"],
```
andrewargatkiny: See line 222.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                            final_layer_params["hidden_size"] // final_layer_params["num_attention_heads"],
+                            num_heads=final_layer_params["num_attention_heads"]
+                        )
+                    else:
+                        rope_cache = relpe_class(
```
andrewargatkiny, line 365: See line 222.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
-                config.relpe_type is not None):
-            self.relpe_type = RelPEType[config.relpe_type.upper()]
+        
+
+
+        if config.layers is not None:
+            rope_caches = []
+            modules = []
+
+            if config.layers_scheme:
```
andrewargatkiny: This outer (lines 137-142) if–elif-else construct code violates DRY principle and could be refactored to use just one execution flow. Generally, it's a good practice to have as little nested layers of conditional flows as possible.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                
+                if config.num_hidden_layers % len(ordered_names) != 0:
+                    raise ValueError("num_hidden_layers must be divisible by the number of layers in layers_scheme.")
+                
+                self.repeats = config.num_hidden_layers // len(ordered_names)
+                
+                self.layer_scheme = []
+                for layer_name in ordered_names:
+                    if layer_name not in name_to_config:
+                        raise ValueError(f"Layer name '{layer_name}' from layers_scheme not found in config.layers.")
```
andrewargatkiny: Please make the message clearer, e.g., "in the config entries describing available layer types".

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+
+                    layer_config_dict = name_to_config[layer_name]
+                    layer_type = layer_config_dict["layer_type"]
+                    
+                    config_class = LayerConfigToClass[layer_type]
+                    final_layer_params = layer_config_dict.copy()
+                    init_signature = inspect.signature(config_class.__init__)
+                    init_params = init_signature.parameters
+
+                    for param_name in init_params:
```
andrewargatkiny: Please add a comment describing what's the purpose of these lines in plain words, so a future reader would understand the logic.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                            final_layer_params["max_position_embeddings"],
+                            final_layer_params["hidden_size"] // final_layer_params["num_attention_heads"],
+                            num_heads=None
+                        )
+                    rope_caches.append(rope_cache)
+
+                    layer_class = LayerTypeToClass[layer_type]
+                    module = layer_class(layer_config)
+                    modules.append(module)
+                modules = modules * self.repeats
```
andrewargatkiny: This brings about a mistake: the number of actual unique layers in the final model would become equal to `len(ordered_names)`. When you repeat a list with mutable objects using "*" operator, only references for original objects get copied, rather than the objects themselves.

andrewargatkiny: In simple words, it means that many layers in the model would share the same tied weights, which is erroneous.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                        final_layer_params["max_position_embeddings"],
+                        final_layer_params["hidden_size"] // final_layer_params["num_attention_heads"],
+                        num_heads=None
+                    )
+                rope_caches.append(rope_cache)
+
+                layer_class = LayerTypeToClass[layer_type]
+                module = layer_class(layer_config)
+                modules.append(module)
+                modules = modules * self.repeats
```
andrewargatkiny: See line 200.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                            num_heads=None
+                        )
+                    rope_caches.append(rope_cache)
+
+                    layer_class = LayerTypeToClass[layer_type]
+                    module = layer_class(layer_config)
+                    modules.append(module)
+                modules = modules * self.repeats
+                self.rope_caches = nn.ModuleList(rope_caches * self.repeats)
+            elif not config.layers_scheme and len(config.layers) == 1:
```
andrewargatkiny: See line 137. The code in this branch could be reformulated to blend with the previous branch.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                        layer_module.attention.self = BertSelfAttention(layer_config)
+                    # 'g' is the default and requires no change, so we just pass.
+            # fallback to old logic for backward compatibility
+            elif config.local_attention:
+                for i, layer in enumerate(self.layer):
+                    if i % 3 == 0:
+                        layer.attention.self = BertSelfLocalAttention(config)
+                    elif i % 3 == 1:
+                        layer.attention.self = BertSelfShiftedLocalAttention(config)
+    def prepare_mask(self, hidden_states, attention_mask, layer_config):
```
andrewargatkiny, line 421: There should be one empty line between function definitions.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                )
+            return extended_attention_mask
+        else:
+            if attention_mask is None:
+                attention_mask = torch.ones_like(hidden_states)
+            attention_mask = attention_mask.to(
+                dtype=hidden_states.dtype)  # fp16 compatibility
+            attention_mask = (1.0 - attention_mask) * -10000.0
+            return attention_mask
     def forward(self,
```
andrewargatkiny, line 449: See line 281.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                        attention_mask / layer_config["window_size"] ** (1. / 3)
+                ).to(dtype)
+                extended_attention_mask = (
+                    local_attention_mask,
+                    extended_attention_mask
+                )
+            return extended_attention_mask
+        else:
+            if attention_mask is None:
+                attention_mask = torch.ones_like(hidden_states)
```
andrewargatkiny: Lines 285-286 and 301-302 are identical in both branches and could be moved outside the `if-else` flow.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                        layer.attention.self = BertSelfLocalAttention(config)
+                    elif i % 3 == 1:
+                        layer.attention.self = BertSelfShiftedLocalAttention(config)
+    def prepare_mask(self, hidden_states, attention_mask, layer_config):
+        
+        if layer_config["layer_type"] == "danet":
+            dtype = hidden_states.dtype
+            if attention_mask is None:
+                attention_mask = torch.ones_like(hidden_states)
+            extended_attention_mask = (
```
andrewargatkiny, line 430: Why does this mask construction differ from the original for DANet?

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                attention_mask /
+                attention_mask.sum(axis=-1, keepdim=True).pow(1. / 3)
+            ).to(dtype)
+            if layer_config["local_scheme"] in ["l", "sl"]:
```
andrewargatkiny: See line 287.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+            # fallback to old logic for backward compatibility
+            elif config.local_attention:
+                for i, layer in enumerate(self.layer):
+                    if i % 3 == 0:
+                        layer.attention.self = BertSelfLocalAttention(config)
+                    elif i % 3 == 1:
+                        layer.attention.self = BertSelfShiftedLocalAttention(config)
+    def prepare_mask(self, hidden_states, attention_mask, layer_config):
+        
+        if layer_config["layer_type"] == "danet":
```
andrewargatkiny, line 427: It seems that current implementation is not identical to the original DANet masking from `src/modeling.py`.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                ).to(dtype)
+                extended_attention_mask = (
+                    local_attention_mask,
+                    extended_attention_mask
+                )
+            return extended_attention_mask
+        else:
+            if attention_mask is None:
+                attention_mask = torch.ones_like(hidden_states)
+            attention_mask = attention_mask.to(
```
andrewargatkiny, line 444: This differs from the original implementation.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                    name_to_config[layer_name] = layer
+
+                ordered_names = config.layers_scheme.split("_")
+                
+                if config.num_hidden_layers % len(ordered_names) != 0:
+                    raise ValueError("num_hidden_layers must be divisible by the number of layers in layers_scheme.")
+                
+                self.repeats = config.num_hidden_layers // len(ordered_names)
+                
+                self.layer_scheme = []
```
andrewargatkiny: Actually, this list holds layer configs rather than layer_scheme (which is just a string). Please rename it accordingly for ease of understanding the subsequent code.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+            return attention_mask
     def forward(self,
                 hidden_states,
                 attention_mask,
-                output_all_encoded_layers=True):
-
+                output_all_encoded_layers=True,
+                **kwargs):
+        masks = []
+        for i in range(len(self.layer_scheme)):
```
andrewargatkiny: It would be better to create an instance variable holding the value of `len(self.layer_scheme)` so it doesn't get recalculated at each forward pass. Please devise a fitting name for it.

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+            attention_mask = (1.0 - attention_mask) * -10000.0
+            return attention_mask
     def forward(self,
                 hidden_states,
                 attention_mask,
-                output_all_encoded_layers=True):
-
+                output_all_encoded_layers=True,
+                **kwargs):
+        masks = []
```
andrewargatkiny: Will this `forward` method work for the legacy case where no `layer_scheme` and/ or `layers` are provided in the config?

src/other_models/modeling.py
andrewargatkiny: Please see new comments.

src/other_models/transformers.py
```diff
@@ -0,0 +1,726 @@
...
+                 local_attention=False,
+                 window_size=1024,
+                 apply_relpe_after=False,
+                 max_position_embeddings=512,
+                 pos_emb_type="learned",
+                 relpe_type=None,
+                 local_scheme=None,
+                 power=2,
+                 scaling_d_factor=False,
+                 **kwargs):
```
andrewargatkiny, line 93: Does this class require **kwargs?

src/other_models/modeling.py
```diff
@@ -645,64 +127,206 @@ def __init__(self, config):
...
+                            final_layer_params["hidden_size"] // final_layer_params["num_attention_heads"],
+                            num_heads=None
+                        )
+                    rope_caches.append(rope_cache)
+
+                    layer_class = LayerTypeToClass[layer_type]
+                    module = layer_class(layer_config)
+                    modules.append(module)
+                modules = modules * self.repeats
+                self.rope_caches = nn.ModuleList(rope_caches * self.repeats)
```
andrewargatkiny: There's no need to `* self.repeats` here, it could lead to subtle errors in the subsequent code.

src/other_models/modeling.py
andrewargatkiny: See new comments.

src/other_models/modeling.py
andrewargatkiny: There are other cases left with `init_bert_weights`.

andrewargatkiny: Please see the new comments.

andrewargatkiny (changes requested):

src/other_models/modeling.py
andrewargatkiny: Still not done according to the new procedure (param `num_heads` is omitted).

src/other_models/modeling.py
```diff
@@ -159,10 +146,17 @@ def __init__(self,
             apply_relpe_after: For linear attentions, determines whether Relative
                 Positional Encoding (RELPE) is applied after the feature map (if true)
                 or before the linear attention kernel (if false).
-            power: For Power Attention, determines the power (p).
+            layers_scheme: Defines the sequence and repetition of layers within the encoder.
+                This should be a string of layer names separated by underscores
+                (e.g., 'layer1_layer2_layer1_layer3'). Each name must correspond to a unique `layer_name`
+                key in one of the configuration dictionaries provided in the layers parameter.
+                power: For Power Attention, determines the power (p).
```
andrewargatkiny: Please correct the tabulation level.

src/other_models/modeling.py
```diff
@@ -159,10 +146,17 @@ def __init__(self,
             apply_relpe_after: For linear attentions, determines whether Relative
                 Positional Encoding (RELPE) is applied after the feature map (if true)
                 or before the linear attention kernel (if false).
-            power: For Power Attention, determines the power (p).
+            layers_scheme: Defines the sequence and repetition of layers within the encoder.
+                This should be a string of layer names separated by underscores
+                (e.g., 'layer1_layer2_layer1_layer3'). Each name must correspond to a unique `layer_name`
+                key in one of the configuration dictionaries provided in the layers parameter.
```
andrewargatkiny: \`layers\`

src/other_models/transformers.py
andrewargatkiny: This problem is not resolved. What's the purpose of this `if else` structure? Regardless of `layers` availability, the `BertEmbeddings` can have its own independent positional encodings.

src/other_models/modeling.py
```diff
@@ -186,7 +180,10 @@ def __init__(self,
             self.attention_probs_dropout_prob = attention_probs_dropout_prob
             self.attn_proj_biases = attn_proj_biases
             self.max_position_embeddings = max_position_embeddings
-            self.pos_emb_type = PositionalEmbeddingsTypes[pos_emb_type.upper()]
+            if layers is None:
```
andrewargatkiny: This problem (I mentioned it earlier) is not resolved. What's the purpose of this `if else` structure? Regardless of `layers` availability, the `BertEmbeddings` can have its own independent positional encodings.

src/danet_layers.py
andrewargatkiny: Still not fixed.

src/other_models/modeling.py
andrewargatkiny: This improvement still can be done.

src/other_models/modeling.py
andrewargatkiny: It's also not resolved.

src/other_models/modeling.py
```diff
@@ -645,65 +295,182 @@ def __init__(self, config):
...
+                init_signature = inspect.signature(config_class.__init__)
+                init_params = init_signature.parameters
+
+                for param_name in init_params:
+                    if param_name not in final_layer_params and hasattr(config, param_name):
+                        final_layer_params[param_name] = getattr(config, param_name)
+                
+                layer_config = config_class(**final_layer_params)
+                
+                if (final_layer_params.get("pos_emb_type") == PositionalEmbeddingsTypes.RELPE and
```
andrewargatkiny: Since you've initialized `layer_config` at this point, lets use it further on for clarity instead of `final_layer_params`.

src/other_models/modeling.py
```diff
@@ -645,65 +295,182 @@ def __init__(self, config):
...
+            elif config.local_attention:
+                for i, layer in enumerate(self.layer):
+                    if i % 3 == 0:
+                        layer.attention.self = BertSelfLocalAttention(config)
+                    elif i % 3 == 1:
+                        layer.attention.self = BertSelfShiftedLocalAttention(config)
+
+    def prepare_mask(self, hidden_states, attention_mask, layer_config):
+        if attention_mask is None:
+            attention_mask = torch.ones(hidden_states.size(0), hidden_states.size(1),
```
andrewargatkiny: Originally, this instantiation proceeded like this: `attention_mask = torch.ones_like(input_ids)`, i.e. attention mask had `int` or `long` dtype (please look it up to be sure which exactly).

src/other_models/modeling.py
```diff
@@ -645,65 +295,182 @@ def __init__(self, config):
...
+            attention_mask = torch.ones(hidden_states.size(0), hidden_states.size(1), 
+                                    dtype=hidden_states.dtype, device=hidden_states.device)
+        if layer_config["layer_type"] == "danet":
+            dtype = hidden_states.dtype
+            
+            extended_attention_mask = (
+                attention_mask /
+                attention_mask.sum(axis=-1, keepdim=True).pow(1. / 3)
+            ).to(dtype).unsqueeze(-1)
+            if layer_config["local_scheme"] in ["l", "sl"]:
```
andrewargatkiny: There are more relevant codes.

src/other_models/modeling.py
```diff
@@ -645,65 +295,182 @@ def __init__(self, config):
...
     def forward(self,
                 hidden_states,
                 attention_mask,
-                output_all_encoded_layers=True):
-
+                output_all_encoded_layers=True,
+                **kwargs):
+        if hasattr(self, "layers_configs"):
+            masks = []
+            num_layers_configs = len(self.layers_configs)
```
andrewargatkiny: Here, it gets calculated at every `forward` call. It can be optimized.

src/other_models/modeling.py
```diff
@@ -645,65 +295,182 @@ def __init__(self, config):
...
+            for i in range(num_layers_configs):
+                masks.append(self.prepare_mask(hidden_states, attention_mask, self.layers_configs[i]))
         all_encoder_layers = []
-        for layer_module in self.layer:
-            hidden_states = layer_module(hidden_states, attention_mask, self.rope_cache)
+        for i, layer_module in enumerate(self.layer):
+            hidden_states = layer_module(
+                hidden_states, 
+                attention_mask=masks[i % num_layers_configs] if hasattr(self, "layers_configs") else attention_mask,
+                rope_cache= self.rope_caches[i % num_layers_configs] if hasattr(self, "rope_caches") else self.rope_cache,
```
andrewargatkiny: `i % num_layers_configs` gets calculated 2 times, could be 1.

src/other_models/modeling.py
```diff
@@ -645,65 +295,182 @@ def __init__(self, config):
...
+                        seq_len=final_layer_params["max_position_embeddings"],
+                        n_elem=final_layer_params["hidden_size"] // final_layer_params["num_attention_heads"],
+                        sep_head_dim=True
+                    )
+                rope_caches.append(rope_cache)
+
+                layer_class = LayerTypeToClass[layer_type]
+                module = layer_class(layer_config)
+                modules.append(module)
+                if i < len(ordered_names):
```
andrewargatkiny, line 348: Let's better move it to immediately after `final_layer_params` completes initialization.

src/other_models/modeling.py
```diff
@@ -645,65 +295,182 @@ def __init__(self, config):
...
+                        n_elem=final_layer_params["hidden_size"] // final_layer_params["num_attention_heads"],
+                        sep_head_dim=False
+                    )
+                else:
+                    rope_cache = relpe_class(
+                        seq_len=final_layer_params["max_position_embeddings"],
+                        n_elem=final_layer_params["hidden_size"] // final_layer_params["num_attention_heads"],
+                        sep_head_dim=True
+                    )
+                rope_caches.append(rope_cache)
```
andrewargatkiny: Number of rope caches could be optimized. Also, you don't need to initialize duplicate layer configs when `i >= len(ordered_names)`.

src/other_models/modeling.py
```diff
@@ -999,40 +766,27 @@ def __init__(self, config):
         self.embeddings = BertEmbeddings(config)
         self.encoder = BertEncoder(config)
```
andrewargatkiny, line 777: Encoder has a FinalLayerNorm param. Is its initialization handled as in the original code?

src/other_models/modeling.py
andrewargatkiny: Still recalculated at each forward pass.

andrewargatkiny (changes requested):

src/other_models/modeling.py
Rozhkin-Aleksandr: Yes.

src/other_models/modeling.py
Rozhkin-Aleksandr: Okay. Now we initialize rope caches only if `i < len(ordered_names)`, but this implementation does not support rope caches with learnable paramaters.

src/other_models/modeling.py
Rozhkin-Aleksandr: Done.

src/other_models/modeling.py
Rozhkin-Aleksandr: Okay, currently `i % num_layers_configs` gets calculated before cycle.

src/other_models/modeling.py
Rozhkin-Aleksandr: Currently, attention mask has `dtype=hidden_states.dtype`.

src/other_models/modeling.py
Rozhkin-Aleksandr: Done.

src/other_models/modeling.py
Rozhkin-Aleksandr: Dropped `if else` structure.

src/other_models/modeling.py
Rozhkin-Aleksandr: Done.

src/danet_layers.py
```diff
@@ -117,3 +117,22 @@ def forward(self, hidden_states, attention_mask, rope_cache):
...
+        std = self.config.initializer_range
+        if isinstance(module, nn.Embedding):
+            module.weight.data.normal_(mean=0, std=std)
+        elif isinstance(module, nn.Linear):
+            module.weight.data.uniform_(-std, std)
+            if module.bias is not None:
+                module.bias.data.zero_()
+
+    def forward(self, hidden_states, attention_mask, rope_cache=None, **kwargs):
+        return self.layer(hidden_states, attention_mask, rope_cache, **kwargs)
```
andrewargatkiny: This line doesn't need `**kwargs`. The wrapper layer can ingest them but inner layer can't.

src/model_config.py
```diff
@@ -51,7 +51,8 @@ def __init__(self,
                  local_scheme="l_sl_g",
                  hybrid=False,
                  transformer_heads=None,
-                 attn_proj_biases=False
+                 attn_proj_biases=False,
+                 **kwargs
```
andrewargatkiny: I don't see a case where this would be required. Can you give an example or restore the original without `**kwargs`?

src/other_models/modeling.py
```diff
@@ -645,65 +292,173 @@ def __init__(self, config):
...
-            self.relpe_type = RelPEType[config.relpe_type.upper()]
+        
+
+
+        if config.layers is not None:
+            rope_caches = []
+            modules = []
+
+            if not config.layers_scheme and len(config.layers) != 1:
+                raise ValueError("Invalid layer configuration. Provide either a 'layers_scheme' with corresponding layer definitions, or a single layer definition in 'config.layers' without a 'layers_scheme'.")
```
andrewargatkiny: The string length is too big, please make it multi-line.

andrewargatkiny: There are other occasions in the new code where the line length is much larger than around 80 symbols. Please find and amend them.

src/other_models/modeling.py
```diff
@@ -645,65 +292,173 @@ def __init__(self, config):
...
+            if not config.layers_scheme:
```
andrewargatkiny: It would be more fitting to move lines 304-305 to before line 317, because the check for existence of `layer_name` in the first layer gets realized later (at line 309).

src/other_models/modeling.py
```diff
@@ -645,65 +292,173 @@ def __init__(self, config):
...
+                config.layers_scheme = config.layers[0]["layer_name"]
+
+            name_to_config = {}
+            for layer in config.layers:
+                if not layer.get("layer_name"):
+                    raise ValueError("Each layer in config.layers must have a 'layer_name' when 'layers_scheme' is used.")
```
andrewargatkiny: We have concurred that it should have a `layer_name` regardless of whether `layers_scheme` is used. So please change the error message.

src/other_models/modeling.py
```diff
@@ -645,65 +292,173 @@ def __init__(self, config):
...
+                            sep_head_dim=False
+                        )
+                    else:
+                        rope_cache = relpe_class(
+                            seq_len=layer_config.max_position_embeddings,
+                            n_elem=layer_config.hidden_size // layer_config.num_attention_heads,
+                            num_heads=layer_config.num_attention_heads,
+                            sep_head_dim=True
+                        )
+                    rope_caches.append(rope_cache)
```
andrewargatkiny: Please add a comment describing limitations of this approach which we have discussed.

src/other_models/modeling.py
```diff
@@ -645,65 +292,173 @@ def __init__(self, config):
...
+                        layer.attention.self = BertSelfLocalAttention(config)
+                    elif i % 3 == 1:
+                        layer.attention.self = BertSelfShiftedLocalAttention(config)
+
+    def prepare_mask(self, hidden_states, attention_mask, layer_config):
+        if attention_mask is None:
+            attention_mask = torch.ones(hidden_states.size(0), hidden_states.size(1), 
+                                    dtype=torch.int32, device=hidden_states.device)
+        if layer_config["layer_type"] == "danet":
+            dtype = torch.int32
```
andrewargatkiny: Wrong data type

src/other_models/modeling.py
```diff
@@ -645,65 +292,173 @@ def __init__(self, config):
...
+            
+            extended_attention_mask = (
+                attention_mask /
+                attention_mask.sum(axis=-1, keepdim=True).pow(1. / 3)
+            ).to(dtype).unsqueeze(-1)
+            if layer_config["local_scheme"] in ["l", "sl", "swa"]:
+                local_attention_mask = (
+                        attention_mask / layer_config["window_size"] ** (1. / 3)
+                ).to(dtype).unsqueeze(-1)
```
andrewargatkiny: Wrong datatype

src/other_models/modeling.py
```diff
@@ -645,65 +292,173 @@ def __init__(self, config):
...
+                extended_attention_mask = (
+                    local_attention_mask,
+                    extended_attention_mask
+                )
+            return extended_attention_mask
+        else:
+            attention_mask = attention_mask.unsqueeze(1).unsqueeze(2)
+            attention_mask = attention_mask.to(
+                dtype=torch.int32)  # fp16 compatibility
```
andrewargatkiny: Wrong datatype

src/other_models/modeling.py
andrewargatkiny: It seems that you've changed all the datatypes in this function mechanistically. Please ensure again that this function **fully** corresponds to original mask preparation procedures for generalized transformer and DANet.

src/other_models/modeling.py
```diff
@@ -645,65 +292,173 @@ def __init__(self, config):
...
+                  **kwargs) 
             if output_all_encoded_layers:
                 all_encoder_layers.append(hidden_states)
         if not output_all_encoded_layers:
             hidden_states = self.FinalLayerNorm(hidden_states)
             all_encoder_layers.append(hidden_states)
 
         return all_encoder_layers
-
+
```
andrewargatkiny: Please remove spaces, let it be just a newline symbol.

src/other_models/modeling.py
andrewargatkiny: Update: I saw the comment on line 452, but you can duplicate it here.

configs/lra/hybrid/dense_attn_pathfinder32_layers.json
andrewargatkiny, line 1: Please move all the new configs to the new `hybrid` subdir and make sure the paths inside the `.sh` script align.

src/other_models/modeling.py
```diff
@@ -645,65 +292,173 @@ def __init__(self, config):
...
+        if attention_mask is None:
+            attention_mask = torch.ones(hidden_states.size(0), hidden_states.size(1), 
+                                    dtype=torch.int32, device=hidden_states.device)
+        if layer_config["layer_type"] == "danet":
+            dtype = torch.int32
+            
+            extended_attention_mask = (
+                attention_mask /
+                attention_mask.sum(axis=-1, keepdim=True).pow(1. / 3)
+            ).to(dtype).unsqueeze(-1)
```
andrewargatkiny, line 433: Wrong datatype

andrewargatkiny (changes requested):

src/danet_layers.py
Rozhkin-Aleksandr: Fixed.

src/other_models/modeling.py
Rozhkin-Aleksandr: Done.

src/other_models/modeling.py
Rozhkin-Aleksandr: Fixed.

src/other_models/modeling.py
Rozhkin-Aleksandr: Fixed.

src/other_models/modeling.py
Rozhkin-Aleksandr: Okay, done.

src/other_models/modeling.py
Rozhkin-Aleksandr: Changed datatype to `hidden_states.dtype`

src/other_models/modeling.py
Rozhkin-Aleksandr: Done.

configs/lra/hybrid/dense_attn_pathfinder32_layers.json
Rozhkin-Aleksandr: Done.

andrewargatkiny:
@codex please review this PR thoroughly, taking into account the whole thread history and comments, as well as initial specs here https://github.com/andrewargatkiny/dense-attention/issues/35

andrewargatkiny:
@codex confirm what you wrote in this comment https://github.com/andrewargatkiny/dense-attention/pull/42#discussion_r2867926141 by launching a minimal script which should invoke this error.

src/other_models/modeling.py
```diff
@@ -645,57 +292,170 @@ def __init__(self, config):
...
+                # Populate missing layer parameters from the main config object
+                init_signature = inspect.signature(config_class.__init__)
+                init_params = init_signature.parameters
+
+                for param_name in init_params:
+                    if param_name not in final_layer_params and hasattr(config, param_name):
+                        final_layer_params[param_name] = getattr(config, param_name)
+
+                layer_class = LayerTypeToClass[layer_type]
+                layer_config = config_class(**final_layer_params)
```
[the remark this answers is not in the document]
andrewargatkiny: Seems like it's a valid concern.

src/other_models/modeling.py
```diff
@@ -645,57 +292,170 @@ def __init__(self, config):
...
-                elif i % 3 == 1:
-                    layer.attention.self = BertSelfShiftedLocalAttention(config)
+            self.rope_cache = RelPETypeToClass[self.relpe_type](
+                seq_len=config.max_position_embeddings,
+                n_elem=config.hidden_size // config.num_attention_heads,
+                sep_head_dim=True
+            )
+            layer = BertLayer(config)
+            self.layer = nn.ModuleList(
+                [copy.deepcopy(layer) for _ in range(config.num_hidden_layers)])
```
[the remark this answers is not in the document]
andrewargatkiny: While this is a valid concern, this behavior is the same as in the original code. So it's going to be saved for now.

src/other_models/modeling.py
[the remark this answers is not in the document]
andrewargatkiny: Fixed.

andrewargatkiny (approved):

PR #43, opened 2025-10-07
Add support for using local datasets with any chunk size

mirrtl:


data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+        files: List of file paths to read from
+        world_size: Total number of shards
+        global_rank: Rank of current shard
+        chunk_size: Size of chunk to read
+        offset: Offset within the shard
+        
+    Returns:
+        List[Dict]: List of data records from the assigned shard
+    """
+    total_len = sum(get_file_len(f) for f in files)
```
andrewargatkiny: Getting lengths for all files in a big dataset (e.g., C4-en has over 1000 chunks  https://huggingface.co/datasets/allenai/c4/tree/main/en) would be manageable only with Parquet datasets having `num_rows` metadata attribute. Otherwise, you'd need to decompress and read each of the thousands of files, at least at the initial pass before the first dataset construction. 

This would involve decompressing and reading up to terabytes of data and cause a very long latency (delay) before the start of the first epoch. 

If we want exact counting of rows, let's do it the following way: read files one at a time until their cumulative row count exceeds desired `offset`. That way, for each new epoch, we have to read at most one data file. And if we resume training, we still have to read and get lengths from a fraction of dataset files, rather than all of them.

andrewargatkiny: Due to the absence of the total dataset length, each of the `global_rank` in `world_size` should have its own independent subset of files to iterate. This would make the algorithm much simpler, but it implies that the number of files should be at least equal to the number of nodes. Please raise an error if the condition is not met.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+        take = min(remaining, file_len - local_offset)
+        if take > 0:
+            data.extend(read_chunk_from_file(path, local_offset, take))
+            remaining -= take
+        file_idx += 1
+        local_offset = 0
+
+    return data
+
+def apply_filters(data: List[Dict], params: LocalDatasetParams) -> List[Dict]:
```
andrewargatkiny: This function would work inefficiently because it uses Python tools (list comprehension) to iterate through the example. Please consider using more optimized methods instead – I suggest `Pandas` or, even better, `Polars` library.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+    for params in params_list:
+        if os.path.isdir(params.path):
+            files = sorted(
+                os.path.join(params.path, f)
+                for f in os.listdir(params.path)
+                if os.path.isfile(os.path.join(params.path, f))
+            )
+        else:
+            files = [params.path]
+
```
andrewargatkiny: There was a requirement in the task description that there should be an option to randomize the order of data (both the order of the data files and the samples within each file).

This line is a good place to permute the order of files if the seed is not None.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+                continue
+            try:
+                obj = json.loads(line)
+            except json.JSONDecodeError:
+                obj = {"text": line}
+            result.append(obj)
+    return result
+
+
+def read_sharded_across_files(files: List[str],
```
andrewargatkiny, line 279: This function should assemble and return an efficient data structure, like `Pandas` or `Polars` DataFrame. Please avoid using plain Python loops and list comprehensions as much as possible – it decrements speed noticeably in comparison with optimized frameworks.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+        pf = pq.ParquetFile(path)
+        n = pf.metadata.num_rows
+    else:
+        n = sum(1 for _ in open_compressed_file(path))
+
+    _file_len_cache[path] = n
+    return n
+
+
+def read_chunk_from_file(path: str, local_offset: int, take: int) -> List[Dict]:
```
andrewargatkiny: Please see the comment on line 320.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+    if path.endswith(".gz"):
+        return gzip.open(path, "rt", encoding="utf-8")
+    if path.endswith(".zst") or path.endswith(".zstd"):
+        dctx = zstd.ZstdDecompressor()
+        f = open(path, "rb")
+        return io.TextIOWrapper(dctx.stream_reader(f), encoding="utf-8")
+    return open(path, "r", encoding="utf-8")
+
+
+def get_file_len(path: str) -> int:
```
andrewargatkiny, line 210: Please make this function pure (i.e. it doesn't depend or have side effects on any variables other than its inputs).

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
             seconds_to_sleep = max(300, seconds_to_sleep + 60)
 
 
+@dataclass
+class LocalDatasetParams:
```
andrewargatkiny: This class is essentially identical to `HFDatasetParams` if we merge `name` and `path` params into a unified `name_or_path` variable. Thus, let's just rename `HFDatasetParams` to `DatasetParams` and that parameter, and stick with one class.

data/dataset_utils.py
andrewargatkiny, line 155: `load_local_dataset` doesn't have this functionality. Please add it.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+    world_size: int = 1
+    global_rank: int = 0
+    dataset_weight: float = 1.0
+    filter_geq_column: Optional[str] = None
+    filter_geq_value: int = 0
+    filter_leq_column: Optional[str] = None
+    filter_leq_value: int = 2**64
+
+
+_file_len_cache: Dict[str, int] = {}
```
andrewargatkiny: The location of this variable and all functions that rely on `get_file_len` poses a problem. `ShardedDatasetWrapper` class uses multiprocessing pool to fetch data for a next epoch concurrently with training on the current epoch – see `prefetch_shard`. It creates new processes with `spawn` method, which loads *everything* from scratch in a new Python Interpreter. The only data which are copied from the parent process rather then instantiated from scratch are arguments supplied to `pool.submit` function.

Here's some info about how multiprocessing works, explained in an accessible manner:

https://britishgeologicalsurvey.github.io/science/python-forking-vs-spawn/

https://pythonspeed.com/articles/python-multiprocessing/

Anticipating a possible question: we can't use `fork` method for creating processes exactly because the framework uses multithreaded frameworks in many places.

In practice, it means that `_file_len_cache` would be instantiated and populated anew at each epoch, making this cache pointless at all.

I see a possible solution in all bookkeeping being made ahead of datasets creation in `ShardedDatasetWrapper` instance. The `dataset_config` being passed into `pool.submit` should already have concrete filename(s) to use and local offset(s) into this/these file(s) inside, ready to be used right away.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+        return io.TextIOWrapper(dctx.stream_reader(f), encoding="utf-8")
+    return open(path, "r", encoding="utf-8")
+
+
+def get_file_len(path: str) -> int:
+    if path in _file_len_cache:
+        return _file_len_cache[path]
+
+    if path.endswith(".parquet"):
+        pf = pq.ParquetFile(path)
```
andrewargatkiny, line 212: Does this create a file handle or file stream? If yes, it would be better wrap it into `with` construct.

data/dataset_utils.py
andrewargatkiny: Additionally, when choosing the right library to load and process raw data, please find out if it can load the required data efficiently, actually skipping the first `local offset` rows, as opposed to initially loading the entire file and discarding the rows afterwards.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+
+    if seed is not None:
+        random.seed(seed)
+
+    datasets = [c for c, _ in chunks_with_weights]
+    weights = [w for _, w in chunks_with_weights]
+    iters = [iter(d) for d in datasets]
+
+    combined = []
+    while datasets:
```
andrewargatkiny: This would be a very slow and inefficient way to assemble a combined dataset. Please design a procedure which uses vectorized or single-call functions.

data/dataset_utils.py
andrewargatkiny: Please also don't forget to implement intra-file randomization.

data/dataset_utils.py
andrewargatkiny: The same applies to line 259.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+
+def get_file_len(path: str) -> int:
+    if path in _file_len_cache:
+        return _file_len_cache[path]
+
+    if path.endswith(".parquet"):
+        pf = pq.ParquetFile(path)
+        n = pf.metadata.num_rows
+    else:
+        n = sum(1 for _ in open_compressed_file(path))
```
andrewargatkiny, line 215: Please investigate whether this works correctly for files with texts which feature `\n` symbols in them. E.g. one line of a jsonl file may contain a text with multiple paragraphs, separated by `\n`, and the `open_compressed_file`  function could erroneously treat it as many lines.

andrewargatkiny: Additionally, perhaps it would be easier and faster to count rows with a call to a library which doesn't suffer from Python overhead for each loop iteration. E.g., Pandas, which automatically reads compressed jsonl files of almost any compression type, or Polars:  https://stackoverflow.com/questions/75523498/python-polars-how-to-get-the-row-count-of-a-lazyframe

However, I'm not totally sure which way is faster and more memory efficient.

data/dataset_utils.py
andrewargatkiny: However, keep in mind that this would calculate the length of a file in the main, parent process, so it must be done in the fastest way possible.

data/dataset_utils.py
```diff
@@ -181,6 +187,262 @@ def materialize_data(ds: datasets.IterableDataset) -> List[str]:
...
+        chunks_with_weights.append((data_chunk, params.dataset_weight))
+
+    if len(chunks_with_weights) == 1:
+        return chunks_with_weights[0][0][:params_list[0].chunk_size]
+
+    if seed is not None:
+        random.seed(seed)
+
+    datasets = [c for c, _ in chunks_with_weights]
+    weights = [w for _, w in chunks_with_weights]
```
andrewargatkiny: This logic for weights initialization does not fully conform to the one specified in `load_hf_datasets`.

andrewargatkiny (changes requested):

andrewargatkiny:
It seems that a simpler solution made in later PRs (#44, #49) works well, so I'm closing this PR for now.

PR #44, opened 2025-10-13
Add support for using local datasets with any chunk size using only datasets library

mirrtl:


andrewargatkiny (changes requested):
There are 2 small problems:

1) When dealing with local datasets: `--data_path_prefix` supplied in `.sh` launch script does not get prepended to the name of the dataset. Please fix it making sure that it doesn't get prepended when cloud dataset from HuggingFace is used (perhaps, you could introduce a new config parameter to signal that we use an offline dataset)

2) In `data/dataset_lm.py` there are 2 occasions of `f"offset {source['offset']}"` in a print statement. If `offset` is omitted in the dataset config, it throws an exception. This one can be fixed straightforwardly.

